PR #31, opened 2024-07-04
feat(ONB-201): Keyword Generator Commit

eececan:
Keyword Generator application is completed.

### DIFF
--- a/src/pages/keyword-generator/KeywordGenerator.vue
+++ b/src/pages/keyword-generator/KeywordGenerator.vue
@@ -1,14 +1,11 @@
 <script setup>
-import { ref } from 'vue';
+import { ref,computed } from 'vue';
 
 const text = ref('');
-const onegrams = ref([]);
-const twograms = ref([]);
-const threegrams = ref([]);
 const keywords = ref([]);
-const showWords = ref(false);
+const maxNGrams = ref(10);
 
-const generate = (keywords, n) => {
+const generateWordArray = (keywords, n) => {
   const result = new Set();
   const words = keywords.split(" ");
   for (let i = 0; i <= words.length - n; i++) {
@@ -18,62 +15,45 @@ const generate = (keywords, n) => {
   return Array.from(result);
 }
 
-const run = () => {
-  if (text.value.trim() === '') {
-    showWords.value = false;
-    keywords.value = [];
-  } else {
-    onegrams.value = generate(text.value, 1);
-    twograms.value = generate(text.value, 2);
-    threegrams.value = generate(text.value, 3);
-    keywords.value = [...onegrams.value, ...twograms.value, ...threegrams.value];
-    showWords.value = true;
+const generateNGrams = () => {
+  keywords.value = [];
+  if (text.value.trim() !== '') {
+    for(let i = 1;i <= maxNGrams.value; i++) {
+      keywords.value.push({i,keywords: generateWordArray(text.value, i)});
+    }
   }
 }
+const validNGrams = computed(()=> {
+  return keywords.value.filter(nGram=>nGram.keywords.length > 0);
+});
 </script>
 
 <template>
   <div class="ma-keywords-generator">
-
     <div class="ma-header">
       <h3>Keyword Generator</h3>
     </div>
 
     <div class="ma-body">
-      <textarea v-model="text" placeholder="Enter a text here..."></textarea>
-      <button @click="run">Generate</button>
+      <div class="text-input">
+        <textarea v-model="text" placeholder="Enter a text here..."></textarea>
+      </div>
+      <button @click="generateNGrams">Generate</button>
     </div>
-    <div class="display" v-if="showWords">
+
+    <div class="display">
       <h3> Generated Keywords </h3>
-      <div v-if="onegrams.length > 0">
-        <h4>One-gram keywords:</h4>
+      <div v-for="nGram in validNGrams" :key="nGram.i">
+        <h4>{{ nGram.i }}-gram keywords:</h4>
         <ul>
-          <li v-for="(keyword, index) in onegrams" :key="'onegram-' + index">{{ keyword }}</li>
+          <li v-for="(keyword,index) in nGram.keywords" :key="nGram.i + index">{{keyword}}</li>
         </ul>
-      </div>
-      <div v-if="twograms.length > 0">
-        <h4>Two-gram keywords:</h4>
-        <ul>
-          <li v-for="(keyword, index) in twograms" :key="'twogram-' + index">{{ keyword }}</li>
-        </ul>
-      </div>
-      <div v-if="threegrams.length > 0">
-        <h4>Three-gram keywords:</h4>
-        <ul>
-          <li v-for="(keyword, index) in threegrams" :key="' threegram-' + index">{{ keyword }}</li>
-        </ul>
-      </div>
-      <div v-else>
-        <p>Nothing entered</p>
       </div>
     </div>
 
   </div>
 </template>
 <style scoped>
-.ma-keywords-generator {
-  /* Add your styles here */
-}
 
 .ma-header {
   font-family: Arial;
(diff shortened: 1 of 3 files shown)


src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,14 +1,11 @@
 <script setup>
-import { ref } from 'vue';
+import { ref,computed } from 'vue';
```
tinas, line 2:
```suggestion
import { ref, computed } from 'vue';
```

tinas: In order to increase the readability of the code, we must take care of its formatting.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,14 +1,11 @@
...
 
 const text = ref('');
-const onegrams = ref([]);
-const twograms = ref([]);
-const threegrams = ref([]);
 const keywords = ref([]);
-const showWords = ref(false);
+const maxNGrams = ref(10);
```
tinas, line 6: Why `maxNGrams` is 10?

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -18,62 +15,45 @@ const generate = (keywords, n) => {
...
-    threegrams.value = generate(text.value, 3);
-    keywords.value = [...onegrams.value, ...twograms.value, ...threegrams.value];
-    showWords.value = true;
+const generateNGrams = () => {
+  keywords.value = [];
+  if (text.value.trim() !== '') {
+    for(let i = 1;i <= maxNGrams.value; i++) {
+      keywords.value.push({i,keywords: generateWordArray(text.value, i)});
+    }
   }
```
tinas, line 24: Here the code only continues if the condition is met. So we can use early return if the condition is not met.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -18,62 +15,45 @@ const generate = (keywords, n) => {
...
 }
+const validNGrams = computed(()=> {
+  return keywords.value.filter(nGram=>nGram.keywords.length > 0);
```
tinas, line 27:
```suggestion
  return keywords.value.filter(nGram => nGram.keywords.length);
```

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -18,62 +15,45 @@ const generate = (keywords, n) => {
...
-          <li v-for="(keyword, index) in onegrams" :key="'onegram-' + index">{{ keyword }}</li>
-        </ul>
-      </div>
-      <div v-if="twograms.length > 0">
-        <h4>Two-gram keywords:</h4>
+      <div v-for="nGram in validNGrams" :key="nGram.i">
+        <h4>{{ nGram.i }}-gram keywords:</h4>
         <ul>
-          <li v-for="(keyword, index) in twograms" :key="'twogram-' + index">{{ keyword }}</li>
+          <li v-for="(keyword,index) in nGram.keywords" :key="nGram.i + index">{{keyword}}</li>
```
tinas, line 49:
```suggestion
          <li v-for="keyword in nGram.keywords" :key="nGram.i + index">{{keyword}}</li>
```

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -18,62 +15,45 @@ const generate = (keywords, n) => {
...
     <div class="ma-header">
       <h3>Keyword Generator</h3>
     </div>
 
     <div class="ma-body">
-      <textarea v-model="text" placeholder="Enter a text here..."></textarea>
-      <button @click="run">Generate</button>
+      <div class="text-input">
+        <textarea v-model="text" placeholder="Enter a text here..."></textarea>
+      </div>
```
tinas, line 40:
```suggestion
      <textarea v-model="text" placeholder="Enter a text here..." class="text-input"></textarea>
```

tinas (changes requested):

src/pages/keyword-generator/KeywordGenerator.vue
eececan: In the task description it was written that the application should show up to 10-gram keywords.

src/pages/keyword-generator/KeywordGenerator.vue
eececan: When I do the suggested change I get the following error: 
" ESLint: Expected 'v-bind:key' directive to use the variables which are defined by the 'v-for' directive. (vue/valid-v-for) "

eececan: However, the application is working as it should. Is this a problem?

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,100 @@
 <script setup>
+import { ref, computed } from 'vue';
+
+const text = ref('');
+const keywords = ref([]);
+
+const generateWordArray = (keywords, n) => {
```
tinas, line 7:
```suggestion
const generateNGram = (keywords, n) => {
```

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,100 @@
...
+    const nGram = words.slice(i, i + n).join(' ');
+    result.add(nGram);
+  }
+  return Array.from(result);
+}
+
+const generateNGrams = () => {
+  keywords.value = [];
+  if (!text.value.trim()) { return; }
+  for(let i = 1;i <= 3; i++) {
```
tinas, line 20:
```suggestion
  for(let i = 1; i <= 3; i++) {
```

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,100 @@
...
+    keywords.value.push({i,keywords: generateWordArray(text.value, i)});
+  }
+}
+const validNGrams = computed(()=> {
+  return keywords.value.filter(nGram => nGram.keywords.length);
+});
```
tinas, line 26: We can pass this computed property directly to the `generateNGrams` function. No need to filter it afterwards.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,100 @@
...
+  font-family: Arial;
+  background-color: #cae3ff;
+  width: 100%;
+  height: 100px;
+  margin-bottom: 10px;
+  border-style: groove;
+  border-radius: 4px;
+}
+
+button {
```
tinas, line 86: We should not style the html element directly.

tinas (changes requested):

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,99 @@
...
+        <h4>{{ nGram.i }}-gram keywords:</h4>
+        <ul>
+          <li v-for="keyword in nGram.keywords" :key="nGram.i + index">{{keyword}}</li>
+        </ul>
+      </div>
+    </div>
+
+  </div>
+</template>
+<style scoped>
```
tinas, line 51: To keep the readability of the code at a high level, we must pay attention to the code layout. We should avoid extra spaces. We should always leave space to the left and right of the operators.

tinas (changes requested):

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,95 @@
 <script setup>
+import { ref } from 'vue';
+const nGrams = ref([]);
```
tinas:
```suggestion
import { ref } from 'vue';

const nGrams = ref([]);
```

tinas (approved):

PR #42, opened 2024-07-10
Improved keyword generator interface initial commit

eececan:


src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,133 @@
...
+};
+
+const generateNGrams = (removeUnwanted) => {
+  if (!text.value.trim()) { return; }
+  const filteredText = removeUnwanted ? removeUnwantedWords() : text.value;
+  nGrams.value = selectedNgrams.value.map(n => ({ n, keywords: generateNGram(filteredText, n) }));
+  nGrams.value = nGrams.value.filter(nGram => nGram.keywords.length);
+};
+
+const removeUnwantedWords = () => {
```
emrekaansurgun, line 40: The removeUnwantedWords function should be a helper function inside generateNGrams, so it should be converted into a function that takes the necessary parameter and works accordingly.

const removeUnwantedWords = (inputText) => {
  const unwantedWords = new Set(['a', 'an', 'the', 'is']);
  return inputText.split(' ').filter(word => !unwantedWords.has(word.toLowerCase())).join(' ');
};

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,133 @@
...
+  const result = new Set();
+  const words = text.split(" ");
+  for (let i = 0; i <= words.length - n; i++) {
+    const nGram = words.slice(i, i + n).join(' ');
+    result.add(nGram);
+  }
+  return Array.from(result);
+};
+
+const generateNGrams = (removeUnwanted) => {
```
emrekaansurgun, line 33: We can also improve this function, and after modifying the removeUnwantedWords function, using it as a helper function in this way will be better.

const generateNGrams = (shouldRemoveCommonWords) => {
  const inputText = text.value.trim();
  if (!inputText) return;

  const filteredText = shouldRemoveCommonWords ? removeCommonWords(inputText) : inputText;
  const generatedNGrams = selectedNgrams.value.map(n => ({
    n,
    keywords: generateNGrams(filteredText, n)
  })).filter(nGram => nGram.keywords.length);

  nGrams.value = generatedNGrams;
};

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,127 @@
...
+  const words = text.split(" ");
+  for (let i = 0; i <= words.length - n; i++) {
+    const nGram = words.slice(i, i + n).join(' ');
+    result.add(nGram);
+  }
+  return Array.from(result);
+};
+
+const generateAllNGrams = (shouldRemoveCommonWords) => {
+  const removeUnwantedWords = (inputText) => {
```
emrekaansurgun, line 33: It would be better to write the removeUnwantedWords function separately and extract it from there.

emrekaansurgun (approved):

PR #59, opened 2024-08-12
 fix(ONB-203): changed places of css imports and added tailwind apply rules

guntastan:


src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -26,33 +26,38 @@
             <template #title>Exclude Words</template>
           </ma-input>
           <div class="multi-select-container">
```
tinas, line 28: For some parts we used the tailwind classes directly, for other parts we used the `@apply` property. For consistency we should choose one of the two methods.

tinas (changes requested):

tinas (approved):

PR #62, opened 2025-07-01
feat(ONB-201): implemented generating keywords functionality

mehmetserhatcelik:
Öncellikle sayfada kullanılcak componentler implement edildi:
Button.vue
TextInput.vue 
Card.vue => Container 
Badge.vue => Keyword'leri gösterebilmek için

KeywordGenerator.vue yukarıdaki componentler kullanılarak yapıldı. Bu aşamada kullanıcı tarafından girilen inputlardan 1,2,3-gram olacak şekilde keywordler çıkartıldı. 

Implement edilen sayfa /keyword-generator adresinde

### DIFF
--- a/src/components/Button.vue
+++ b/src/components/Button.vue
@@ -1,7 +1,32 @@
+<script setup>
+
+const emit = defineEmits(['click']);
+
+defineProps({
+  icon: {
+    type: [Object],
+    default: null
+  },
+  label: {
+    type: String,
+    default: ''
+  },
+  disabled: {
+    type: Boolean,
+    default: false
+  }
+});
+
+function handleClick(event) {
+  emit('click', event);
+}
+
+</script>
+
 <template>
   <button
     class="ma-button"
-    @click="$emit('click', $event)"
+    @click="handleClick"
     type="button"
     :disabled="disabled"
   >
@@ -15,34 +40,36 @@
   </button>
 </template>
 
-<script setup>
-defineProps({
-  icon: {
-    type: [Object, Function, String],
-    default: null
-  },
-  label: {
-    type: String,
-    default: ''
-  },
-  disabled: {
-    type: Boolean,
-    default: false
+<style lang="scss" scoped>
+  .ma-button {
+      /* @apply inline-flex items-center gap-2 px-3 py-2 border rounded-xl hover:bg-gray-100 transition-colors font-semibold text-sm w-fit; */
+      display: inline-flex;
+      align-items: center;
+      gap: 8px;
+      padding: 8px 12px;
+      border: 1px solid #e5e7eb;
+      border-radius: 12px;
+      font-weight: 600;
+      font-size: 14px; 
+      width: fit-content;
+      transition-property: background-color;
+      transition-duration: 150ms;
   }
-});
-
-defineEmits(['click']); 
-
-</script>
-
-<style scoped>
-.ma-button {
-    @apply inline-flex items-center gap-2 px-3 py-2 border rounded-xl hover:bg-gray-100 transition-colors font-semibold text-sm w-fit;
-}
-.ma-button:disabled {
-    @apply opacity-50 cursor-not-allowed;
-}
-.ma-button-icon {
-    @apply flex items-center w-4 h-4 text-sm font-semibold; 
-}
+  .ma-button:hover {
+      background-color: #f3f4f6;
+  }
+  .ma-button:disabled {
+      /* @apply opacity-50 cursor-not-allowed; */
+      opacity: 0.5;
+      cursor: not-allowed;
+  }
+  .ma-button-icon {
+      /* @apply flex items-center w-4 h-4 text-sm font-semibold; */
+      display: flex;
+      align-items: center;
+      width: 16px;
+      height: 16px; 
+      font-size: 14px; 
+      font-weight: 600;
+  }
 </style>
--- a/src/components/Button.vue
+++ b/src/components/Button.vue
@@ -1,10 +1,7 @@
 <script setup>
-
-const emit = defineEmits(['click']);
-
 defineProps({
   icon: {
-    type: [Object],
+    type: Object,
     default: null
   },
   label: {
@@ -16,6 +13,8 @@ defineProps({
     default: false
   }
 });
+
+const emit = defineEmits(['click']);
 
 function handleClick(event) {
   emit('click', event);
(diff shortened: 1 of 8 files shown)


sevvalkaplannn (changes requested):
Screenshot 2025-07-02 at 10 06 35
Hi serhat, before reviewing your code, there is a feature that is wrong on the page, you should not duplicate the same word when printing 1gram words, the sample example is in the documentation, can you fix this?


https://mobileaction.atlassian.net/wiki/spaces/Engineerin/pages/1456243369/Implement+Your+First+Task+-+FrontEnd


And can you make a home page with a button that directs to the relevant page so that we don't have to go through the URL?

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,116 @@
...
+import Card from '@/components/Card.vue';
+import TextInput from '@/components/TextInput.vue';
+import Button from '@/components/Button.vue';
+import { ArrowPathIcon } from '@heroicons/vue/24/outline';
+import Badge from '@/components/Badge.vue';
+
+const input = ref('');
+const n_grams = ref([1, 2, 3]);
+
+function getNGrams(words, n) {
```
sevvalkaplannn, line 12: in the script setup part there is a rule that you should follow for implementation order for example your function is written before the below computed, it should be below; 

1- Props and Emits
2- Ref and Reactive data
3- Computed property
4-Method
5-Function description
6- Lifecycle hooks

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,116 @@
...
+  return Array.from(new Set(result));
+}
+
+const ngramsResult = computed(() => {
+  let words = input.value
+    .split(/\s+/)
+    .map(w => w.trim().toLowerCase())
+    .filter(Boolean);
+
+  words = words.filter((w, i, arr) => i === 0 || w !== arr[i - 1]);
```
sevvalkaplannn, line 27: you don't need === or !== you can use == and != since you are not checking the types, === is used for type checking, since we're comparing a number with a number literal, and strings with strings you don't need it

sevvalkaplannn (changes requested):
I added notes for code changes, but there are two additional comments, 

1- when there is no keyword in the input it still shows that there is 1 words, 
Screenshot 2025-07-02 at 11 08 41

2- The feature to filter duplicates is not working properly. In the example in the image, the output should be fox fox and there there in 2-grams, and fox fox fox should be in 3-grams. What is meant here is not to write the same n-gram twice in the outputs.

Screenshot 2025-07-02 at 11 09 26

src/components/Button.vue
```diff
@@ -1,24 +1,10 @@
...
-    <span v-if="label">
-      {{ label }}
-    </span>
-    <slot v-else />
-  </button>
-</template>
-
 <script setup>
+
+const emit = defineEmits(['click']);
```
sevvalkaplannn, line 3: it can be under the props.

src/components/TextInput.vue
```diff
@@ -1,13 +1,6 @@
...
-    <textarea
-        :type="type"
-        :placeholder="placeholder"
-        :value="modelValue"
-        :class="['ma-text-input', customClass]"
-        @input="$emit('update:modelValue', $event.target.value)"
-    />
-</template>
 <script setup>
+// 1- Props and Emi
```
sevvalkaplannn, line 2: you can delete it

src/components/TextInput.vue
```diff
@@ -1,13 +1,6 @@
...
+const emit = defineEmits(['update:modelValue']);
```
sevvalkaplannn, line 3: it can be under the props.

src/pages/index.vue
```diff
@@ -1,10 +1,16 @@
 <script setup>
+import Button from '@/components/Button.vue'
+import { useRouter } from 'vue-router'
 
+const router = useRouter();
```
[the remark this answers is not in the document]
sevvalkaplannn: for good code readability you should add an empty line between code blocks if they are doing different things, for example if you have const's like 

const a = useRouter();
const b = use...();

you don't need to add a blank line between them

but in this case you have function below the const, so you can add a blank line, it is not a rule but it looks good and professional when we do it like that :)

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,116 @@
...
-            <span>Keyword Generator</span>
+            <h1 class="ma-title">Keyword Generator</h1>
+            <p class="ma-description">Generate n-gram keywords from your text. Perfect for app descriptions, and ASO analysis.</p>
         </div>
+        <Card title="Text Input" description="Enter your text to generate keywords. For example, paste an app description from the App Store.">
+            <div class="ma-text-input-container">
+                <TextInput v-model="input"/>
+            </div>
+            <div class="ma-button-container">
+                <div class="ma-description-count">{{ input.length }} characters, {{ input.split(/\s+/).length }} words</div>
```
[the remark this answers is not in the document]
sevvalkaplannn: I left a comment into your implementation

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -33,10 +23,29 @@ const ngramsResult = computed(() => {
...
     input.value = '';
 }
+
+function useTextStats(text) {
+  const charCount = computed(() => text.value.trim().length);
+  const wordCount = computed(() => text.value.trim().split(/\s+/).filter(Boolean).length);
+  return { charCount, wordCount };
+}
+
+const { charCount, wordCount } = useTextStats(input);
```
sevvalkaplannn, line 46: you don't need them, you can make the composable like below `const inputInfo = computed(() => input.value.length + ' characters, ' + input.value.split(/\s+/).length + ' words');
` and then you can simply use it by calling like {{ inputInfo }} in the template part.

src/components/Button.vue
```diff
@@ -1,24 +1,10 @@
...
+
 defineProps({
   icon: {
-    type: [Object, Function, String],
+    type: [Object],
```
sevvalkaplannn, line 7: you don't need to put the Object in an array type: Object, will be enough

sevvalkaplannn (changes requested):

PR #64, opened 2025-07-02
feat(ONB-201): Keyword Generator implemented

ufukagaya:
- Implemented Keyword Generator page at address /keyword-generator.
- Keywords are sorted according to their frequency of use in the app description and distinguished with colors.
- Keywords clicked on are highlighted in the app description provided.

### DIFF
--- a/src/components/TheNav.vue
+++ b/src/components/TheNav.vue
@@ -1,7 +1,7 @@
 <template>
     <div class="ma-router-item">
-        <router-link to="/keyword-density">
-            Keyword Density
+        <router-link to="/keyword-generator">
+            Keyword Generator
         </router-link>
     </div>
 </template>
--- a/src/pages/keyword-generator/KeywordGenerator.vue
+++ b/src/pages/keyword-generator/KeywordGenerator.vue
@@ -1,9 +1,166 @@
 <script setup>
+import { ref, computed } from 'vue';
+import { cleanDescription } from '../../utils/CleanDescription';
+import { filterArr } from '../../cleanupResources';
+
+const userInput = ref('');
+const generatedKeywords = ref(null);
+const activeTab = ref('1-gram');
+const selectedKeyword = ref(null);
+
+function generateKeywords() {
+    const cleaned = cleanDescription(userInput.value);
+    const words = cleaned.split(/\s+/).filter(word => word && !filterArr.includes(word));
+
+    generatedKeywords.value = {
+        '1-gram': generateNGrams(words, 1),
+        '2-gram': generateNGrams(words, 2),
+        '3-gram': generateNGrams(words, 3),
+    };
+    selectedKeyword.value = null;
+}
+
+function generateNGrams(words, n) {
+    const ngrams = new Map();
+    for (let i = 0; i <= words.length - n; i++) {
+        const ngram = words.slice(i, i + n).join(' ');
+        ngrams.set(ngram, (ngrams.get(ngram) || 0) + 1);
+    }
+    const sortedNgrams = Array.from(ngrams.entries())
+        .sort(([, countA], [, countB]) => countB - countA)
+        .map(([keyword, count]) => ({ keyword, count }));
+    return sortedNgrams;
+}
+
+function clearInput() {
+    userInput.value = '';
+    generatedKeywords.value = null;
+    selectedKeyword.value = null;
+}
+
+const highlightedDescription = computed(() => {
+    if (!userInput.value) return '';
+    if (!selectedKeyword.value) {
+        return userInput.value.replace(/[&<>"'`]/g, char => ({ '&': '&amp;', '<': '&lt;', '>': '&gt;', '"': '&quot;', "'": '&#39;', '`': '&#x60;' }[char]));
+    }
+
+    // Handle punctuations
+    const words = selectedKeyword.value.split(' ').map(word => word.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'));
+    const regex = new RegExp(`(${words.join('\\W+')})`, 'gi');
+
+    return userInput.value.replace(regex, `<span class="highlight">$1</span>`);
+});
+
+const currentKeywords = computed(() => {
+    return generatedKeywords.value ? generatedKeywords.value[activeTab.value] : [];
+});
+
+const maxCount = computed(() => {
+    if (!currentKeywords.value.length) return 0;
+    return Math.max(...currentKeywords.value.map(k => k.count));
+});
+
+function getKeywordStyle(count) {
+    const colors = ['#FF0000', '#FFA500', '#FFFF00', '#008000', '#00FFFF', '#0000FF', '#4B0082', '#480048'];
+    const size = 1 + (count / maxCount.value) * 0.5;
+
+    let colorIndex = 0;
+    if (maxCount.value > 1) {
+        colorIndex = Math.round(((count - 1) / (maxCount.value - 1)) * (colors.length - 1));
+    }
+
+    return {
+        fontSize: `${size}em`,
+        color: colors[colorIndex],
+    };
+}
 </script>
+
 <template>
-    <div class="ma-keywords-generator">
-        <div class="ma-header">
-            <span>Keyword Generator</span>
+    <div class="container mt-5">
+        <div class="row">
+            <div class="col-md-5">
+                <div class="card">
+                    <div class="card-body">
+                        <h5 class="card-title">Keyword Generator</h5>
+                        <p class="card-text">Paste your app description to extract keywords.</p>
+                        
+                        <div v-if="generatedKeywords" class="description-area" v-html="highlightedDescription"></div>
+                        <textarea v-else v-model="userInput" class="form-control" rows="8" placeholder="Paste your app description here..."></textarea>
+
+                        <div class="mt-3">
+                            <button @click="generateKeywords" class="btn btn-primary me-2">Generate Keywords</button>
+                            <button @click="clearInput" class="btn btn-secondary">Clear</button>
+                        </div>
+                    </div>
+                </div>
+            </div>
+            <div class="col-md-7">
+                <div v-if="generatedKeywords" class="card">
+                    <div class="card-header">
+                        <ul class="nav nav-tabs card-header-tabs">
+                            <li class="nav-item" v-for="(keywords, ngram) in generatedKeywords" :key="ngram">
+                                <a class="nav-link" :class="{ active: activeTab === ngram }" @click="activeTab = ngram; selectedKeyword = null;" href="#">
+                                    {{ ngram }} ({{ keywords.length }})
+                                </a>
+                            </li>
+                        </ul>
+                    </div>
+                    <div class="card-body">
+                        <div v-if="currentKeywords.length">
+                            <div class="keywords-container">
+                                <span v-for="item in currentKeywords" :key="item.keyword" class="badge m-1 p-2" :style="getKeywordStyle(item.count)" @click="selectedKeyword = item.keyword">
+                                    {{ item.keyword }} ({{ item.count }})
+                                </span>
+                            </div>
+                        </div>
+                        <div v-else>
+                            <p class="text-center text-muted">No keywords found for this n-gram.</p>
+                        </div>
+                    </div>
+                </div>
+                <div v-else class="d-flex justify-content-center align-items-center h-100">
+                    <p class="text-muted">Keywords will appear here once generated.</p>
+                </div>
+            </div>
         </div>
     </div>
 </template>
+
+<style scoped>
+.keywords-container {
+    display: flex;
+    flex-wrap: wrap;
+    align-items: center;
+}
+.badge {
+    transition: all 0.2s ease-in-out;
+    cursor: pointer;
+    border: 1px solid transparent;
+}
+.badge:hover {
+    transform: scale(1.05);
+    border: 1px solid #ccc;
+}
+.nav-link {
+    cursor: pointer;
+}
+.description-area {
+    white-space: pre-wrap;
+    padding: 0.375rem 0.75rem;
+    border: 1px solid #ced4da;
+    border-radius: 0.25rem;
+    margin-bottom: 1rem;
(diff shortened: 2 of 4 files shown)


sevvalkaplannn (changes requested):
https://github.com/user-attachments/assets/bdd2219e-75b4-42b7-b590-95c8b72cbc68

hi ufuk, 

1- when we generate the keywords in the text input we are not able to edit the input unless we click the clear button, it is a way that doing it so it is not wrong way to do it, but for the user experience it is not the best way to do it, we should be able to delete or edit the text input whenever we want, we shouldn't need to click the clear button, because maybe user wanna add a new keyword or made a typo but now when we want to make changes we need to clear all of the text, it is not proper for input component. Can you change it?

2- can you make a home page with a button that directs to the relevant page so that we don't have to go through the URL for keyword generator page?

3- you shouldn't use bootstrap for css, you should use css flexbox, for that you can read those documentations

- https://www.w3schools.com/css/css3_flexbox.asp
- https://css-tricks.com/snippets/css/a-guide-to-flexbox/

sevvalkaplannn (changes requested):
ufuk hi, you shouldn't add the node modules into the pr, please go to the master and copy the .gitignore and paste it in to yours

sevvalkaplannn (changes requested):
hi ufuk

1- you don't need to make two displays for input, below input is enough for our case, but the css is not aligned. Can you write it with better alignments, the input overflows from your borders. Please check the example input in the documentation.

Screenshot 2025-07-02 at 23 25 53

2- when writing css classes you need to name it better, for example for .home-button class you should name it like ma-home-button, you should add ma-.. for the naming convention (ma means mobileaction for writing company based code), and the names of the CSS classes should be compatible with each other and when we see the class name, it should give information about where it is used.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,254 @@
 <script setup>
```
sevvalkaplannn, line 1: in the script setup part there is a rule that you should follow for implementation order it should follow the below ruleset;

1- Props and Emits
2- Ref and Reactive data
3- Computed property
4-Method
5-Function description
6- Lifecycle hooks

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,254 @@
...
+    position: absolute;
+    top: 1rem;
+    left: 1rem;
+    text-decoration: none;
+}
+.ma-container {
+    margin-top: 5rem;
+    max-width: 100vw;
+    overflow-x: hidden;
+}
```
sevvalkaplannn, line 122: Can you wrap the CSS classes within each other according to the hierarchy between the divs, like the example below, this will increase readability. 

`
.ma-container {
    margin-top: 5rem;
    max-width: 100vw;
    overflow-x: hidden;
    .ma-home-button {
       position: absolute;
       top: 1rem;
       left: 1rem;
       text-decoration: none;
    }
}`

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,254 @@
...
+.ma-btn {
+    display: inline-block;
+    font-weight: 400;
+    text-align: center;
+    white-space: nowrap;
+    vertical-align: middle;
+    -webkit-user-select: none;
+    -moz-user-select: none;
+    -ms-user-select: none;
+    user-select: none;
```
sevvalkaplannn, line 182: I think you are writing unnecessary css's can you check the css's and delete the ones that you do not need.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,254 @@
...
+}
+.ma-badge {
+    display: inline-block;
+    padding: 0.25em 0.4em;
+    font-size: 0.95em;
+    font-weight: 600;
+    line-height: 1;
+    text-align: center;
+    white-space: nowrap;
+    vertical-align: baseline;
```
sevvalkaplannn, line 238: for example you don't need them

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,254 @@
...
 <template>
-    <div class="ma-keywords-generator">
-        <div class="ma-header">
-            <span>Keyword Generator</span>
+    <div class="ma-container">
+        <router-link to="/" class="ma-btn ma-btn-primary ma-home-button">Home</router-link>
+        <div class="ma-row">
+            <div class="ma-col-left">
+                <div class="ma-card">
+                    <div class="ma-card-body">
```
sevvalkaplannn, line 61: you are wrapping the same code block with 3 different div's, you don't need them you can just have one and you can collect the css's in one class.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,254 @@
...
+                            <li class="ma-nav-item" v-for="(keywords, ngram) in generatedKeywords" :key="ngram">
+                                <a class="ma-nav-link" :class="{ active: activeTab === ngram }" @click="activeTab = ngram" href="#">
+                                    {{ ngram }} ({{ keywords.length }})
+                                </a>
+                            </li>
+                        </ul>
+                    </div>
+                    <div class="ma-card-body">
+                        <div v-if="currentKeywords.length">
+                            <div class="ma-keywords-container">
```
sevvalkaplannn, line 92: in here too, you don't need to wrap the same code block with two different divs. can you check the whole code for removing the unnecessary duplicated divs.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,254 @@
...
+});
+
 </script>
+
 <template>
-    <div class="ma-keywords-generator">
-        <div class="ma-header">
-            <span>Keyword Generator</span>
+    <div class="ma-container">
+        <router-link to="/" class="ma-btn ma-btn-primary ma-home-button">Home</router-link>
```
sevvalkaplannn, line 57: you already have src/components/TheNav.vue component, you can use that instead of writing this part.

src/components/TheNav.vue
```diff
@@ -1,16 +1,23 @@
...
+}
+.ma-nav-link {
+    text-decoration: none;
+    color: #007bff;
+    padding: 0.5rem 1rem;
+}
+.ma-nav-link:hover {
+    text-decoration: underline;
+}
 </style>
```
sevvalkaplannn, line 23: you updated the TheNav.vue component but you are not using it anywhere

src/App.vue
```diff
@@ -1,13 +1,33 @@
...
+    display: flex;
+    justify-content: center;
+    align-items: flex-start;
+    padding: 20px;
+}
+.ma-page-wrapper {
+    width: 100%;
+    max-width: 1200px;
+}
+</style>
```
sevvalkaplannn, line 33: You can wrap the CSS classes within each other according to the hierarchy between the divs, like the example below, this will increase readability

`.ma-app-container {
    .....
    .ma-main-content {
        .....
        .ma-page-wrapper {
           ......
        }
     }
}
`

src/pages/index.vue
```diff
@@ -1,10 +1,52 @@
...
+                <router-link to="/keyword-generator" class="ma-btn ma-btn-primary">Go to Keyword Generator</router-link>
+            </div>
+        </div>
     </div>
-</template>
\ No newline at end of file
+</template>
+
+<style scoped>
+.ma-home-container {
```
sevvalkaplannn, line 13: please follow the instruction that I wrote about the css class hierarchy in the above and below comments.

src/pages/index.vue
```diff
@@ -1,10 +1,52 @@
...
+}
+.ma-btn {
+    display: inline-block;
+    font-weight: 400;
+    text-align: center;
+    white-space: nowrap;
+    vertical-align: middle;
+    -webkit-user-select: none;
+    -moz-user-select: none;
+    -ms-user-select: none;
```
sevvalkaplannn, line 37: do you need them? please check the whole css codes and delete the unused ones.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,254 @@
...
+    nextTick(() => {
+        if (textareaRef.value) {
+            textareaRef.value.style.height = 'auto';
+            textareaRef.value.style.height = textareaRef.value.scrollHeight + 'px';
+        }
+    });
+}
+
+onMounted(autoResize);
+watch(userInput, autoResize);
```
sevvalkaplannn, line 20: you don't need that function, you can just give a min-height into your input class, it will visible up to the min height limit, then there is no problem with scrolling. You don't need to write js codes for that case, just give min-height

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,254 @@
...
+                    <div class="ma-card-body">
+                        <h5 class="ma-card-title">Keyword Generator</h5>
+                        <textarea
+                            v-model="userInput"
+                            class="ma-textarea"
+                            rows="2"
+                            placeholder="Paste your app description here..."
+                            ref="textareaRef"
+                            @input="autoResize"
+                            style="resize:none;"
```
sevvalkaplannn, line 70: you don't need to add style in here, in your css classes you can write them

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,254 @@
...
+function generateKeywords() {
+    const cleaned = cleanDescription(userInput.value);
+    const words = cleaned.split(/\s+/).filter(word => word);
+
+    generatedKeywords.value = {
+        '1-gram': generateNGrams(words, 1),
+        '2-gram': generateNGrams(words, 2),
+        '3-gram': generateNGrams(words, 3),
+    };
+}
```
sevvalkaplannn, line 31: can you get the words then map them according their n-grams, instead of doing it for 3 of them.

sevvalkaplannn (changes requested):

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,254 @@
...
+                </div>
+            </div>
+            <div class="ma-col-right">
+                <div v-if="generatedKeywords" class="ma-card">
+                    <div class="ma-card-header">
+                        <ul class="ma-nav">
+                            <li class="ma-nav-item" v-for="(keywords, ngram) in generatedKeywords" :key="ngram">
+                                <a class="ma-nav-link" :class="{ active: activeTab === ngram }" @click="activeTab = ngram" href="#">
+                                    {{ ngram }} ({{ keywords.length }})
+                                </a>
```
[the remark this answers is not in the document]
sevvalkaplannn: for example this is a better feedback, I missed that.

ufukagaya:
- Refactored and simplified the KeywordGenerator.vue component:
- Removed unnecessary wrapper divs and redundant classes.
- Ensured all class names use the ma- prefix and are written in a single line in CSS.
- Moved button styles  from the global CSS file into the component’s scoped style block.
- Updated the <script setup> section to follow a clear order: refs, computed, methods, helpers.
- Optimized n-gram generation to process all n-grams in a single loop.
- Ensured cleanDescription uses utilities from cleanupResources.js.
- Removed all JavaScript related to textarea auto-resizing; now uses only CSS with min-height and overflow-y: auto.
-Updated App.vue, index.vue, TheNav.vue components to use <style lang="scss" scoped> and single-line class definitions for consistency.

sevvalkaplannn (changes requested):
Screenshot 2025-07-03 at 16 57 28

Screenshot 2025-07-03 at 16 57 48

you have errors on your src/pages/keyword-generator/KeywordGenerator.vue file ufuk can you fix them? 

and in your own ide doesn't it show you those errors? 

2- We misunderstood about the feedback about css classes, I did not mention that the css codes should be single-line, (it will be written as single-line when using tailwind in the later stages, but not for plain css). What I wanted to say is for example in your src/pages/keyword-generator/KeywordGenerator.vue file you have div which has .ma-container class, and it wraps the div with ma-card-content class, so in your css you can write them nested too. I showed you in below picture for just one. 

Screenshot 2025-07-03 at 17 05 08

Screenshot 2025-07-03 at 17 04 51

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,211 @@
 <script setup>
+import { ref, computed } from 'vue';
+import { regex, splitRegex } from '../../cleanupResources.js';
```
sevvalkaplannn, line 3: ufuk in src/utils/CleanDescription.js we are using it in a function, you should call that function, not them

sevvalkaplannn (changes requested):

ufukagaya:
- Refactored Nav.vue usage.
- Deleted and simplified unnecessary classes.

ufukagaya:
- Removed style usage.
- Improved button visions

src/components/TheNav.vue
```diff
@@ -1,16 +1,20 @@
...
+}
+</script>
+
 <template>
-    <div class="ma-router-item">
-        <router-link to="/keyword-density">
-            Keyword Density
-        </router-link>
-    </div>
+    <button class="ma-btn ma-btn-secondary ma-home-fixed" @click="goHome">Home</button>
```
sevvalkaplannn, line 10: you are not using the extra class in here
```suggestion
    <button class="ma-home-fixed" @click="goHome">Home</button>
```

src/components/TheNav.vue
```diff
@@ -1,16 +1,20 @@
...
 </template>
 
-<style>
- .ma-router-item {
-   display: flex;
-   flex-direction: row;
-   justify-content: space-evenly;
-   margin: 15px;
- }
+<style scoped>
```
sevvalkaplannn, line 13:
```suggestion
<style lang="scss" scoped>
```

PR #71, opened 2025-07-08
feat(ONB-203): implement keyword density analysis feature with UI enh…

mehmetserhatcelik:
…ancements

### DIFF
--- a/src/pages/keyword-density/KeywordDensity.vue
+++ b/src/pages/keyword-density/KeywordDensity.vue
@@ -1,7 +1,8 @@
 <script setup>
 import { ref, computed, onMounted } from 'vue';
 import { MaInput, MaButton, MaCard, MaNotification, MaTooltip2 as MaTooltip, MaBadge, MaEmpty, MaCheckbox2 as MaCheckbox, MaPagination } from '@mobileaction/action-kit'
-import { cleanDescription, cleanStopWords } from '@/utils/CleanDescription';
+import { cleanDescription } from '@/utils/CleanDescription';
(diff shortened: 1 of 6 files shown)


src/pages/keyword-density/KeywordDensity.vue
```diff
@@ -1,21 +1,391 @@
...
+
+const cleanedInput = computed(() => cleanDescription(input.value).trim().toLowerCase().split(/\s+/).filter(Boolean));
+
+const inputInfo = computed(() => {
+  const charCount = input.value.trim().length;
+  const wordCount = cleanedInput.value.length;
+  return charCount +' characters, '+ wordCount +' words'
+})
+
+function analyzeKeywords(showNotification = true) {
```
sevvalkaplannn, line 40: your analyzeKeywords function seems very complex, and it is not clearly readable, can you optimize it. Maybe you can split it into composables.

sevvalkaplannn (changes requested):
1- when I'm in the second page in the table, and then I change the text which has 1 page data, the table looks buggy, it should be dynamically settled to first page when the text changes.

https://github.com/user-attachments/assets/151537f0-c7f6-46cf-971a-c94e253923e6

2- When there are multiple keywords that have same density your Analysis Statistics are misleading us, for example in this case lorem and sit have same density but I only see the lorem in your statistics. If we are making that feature, it should work for every cases.

Screenshot 2025-07-09 at 15 31 26

sevvalkaplannn (changes requested):
can you make a composable utility file and move those functions?

sevvalkaplannn (approved):
good job serhat

PR #73, opened 2025-07-16
feat(ONB-204): integrate Pinia for state management in keyword densit…

mehmetserhatcelik:
…y and generator pages


sevvalkaplannn (changes requested):
this store construction is an option, but would you write this composition setup like regular js composables, without state, getters and actions, use classic methods and use refs intead of state part

PR #74, opened 2025-07-16
feat(ONB-203): Convert CSS classes to Tailwind.

ufukagaya:
- Added an action-kit for converting user text to keywords.
- Added an icon for the button.


sevvalkaplannn (changes requested):
It still can be generated via n-gram buttons and filter unwanted keywords checkbox. We made the styling for attracting the user to generate button, the results should only change when we click the button. 
https://github.com/user-attachments/assets/ba6a91bd-88a4-42af-9664-64c5f95606af

PR #77, opened 2025-07-17
feat(ONB-204): Implement Keyword Density Page.

ufukagaya:
- Implemented Keyword Density page.
- Implemented MaNatification and copy to clipboard button.
- Refactored workflow with data export functionality.

### DIFF
--- a/src/components/TheNav.vue
+++ b/src/components/TheNav.vue
@@ -6,6 +6,7 @@ import { computed } from 'vue';
 const PAGES = {
     HOME: '/',
     KEYWORD_GENERATOR: '/keyword-generator',
+    KEYWORD_DENSITY: '/keyword-density',
 };
 
 const router = useRouter();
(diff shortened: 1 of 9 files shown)


src/components/TheNav.vue
```diff
@@ -17,47 +18,57 @@ function goHome() {
...
-const useNavButtonVariant = (pagePath) => {
-    return computed(() => route.path == pagePath ? 'filled' : 'ghost');
-};
-
-const useNavButtonColor = (pagePath) => {
-    return computed(() => route.path == pagePath ? 'dark' : 'primary');
+const useNavButtonStyle = (pagePath) => {
+    return computed(() => ({
+        variant: route.path === pagePath ? 'filled' : 'ghost',
+        color: route.path === pagePath ? 'dark' : 'primary'
```
sevvalkaplannn, line 28: 🥲please do not forget old feedbacks

sevvalkaplannn (changes requested):
task has a point that "It will get static ‘text’ value from parent page for first load." in your project there is no initial text that comes when we enter the keyword density page, can you implement it, and in your keyword generator page you have filter unwanted keywords checkbox, but you are automatically filtering the unwanted keywords, I think you should add that checkbox into keyword density file too.

sevvalkaplannn (changes requested):
1- can you add total characters like in the example visualization in the task
2- page should open with static texts results, now we have to click the button, but keywords density's first render we should see the initial text's results on the table.

ufukagaya:
- A total character counter has been added to the Keyword Density page.
- Set the page to open with the results of the static mock text.
- Added an 'Analyze Density' button to the Keyword Generator page to transfer the text to the Keyword Density page.
- Changed the UI of the buttons for a better user experience.

sevvalkaplannn (changes requested):
1- Analyze density button is not working, it looks like it's buggy because of the initial text, if you wanna keep the analyze density button, please add MaTooltip for that button because user should know it's purpose, and if you are keeping it, when we click that button we should be redirected to keyword density page with that text, it is more ux friendly.

https://github.com/user-attachments/assets/f0f27b17-cf53-45c4-913e-2a709efd4ca1

ufukagaya:
I see the issue. After final refactoring, i tried to use malinkbutton instead of mabutton with router. I noticed this change caused this bug and fixed it. I have been waiting for review for a new commit.

PR #78, opened 2025-07-21
feat(ONB-205): Add Pinia store for state management.

ufukagaya:
- Create keywordAnalysis store with all required state and actions.
- Migrate both pages from local state to Pinia store-based state management.
- Add computed properties for text stats, word frequency, and pagination.
- Implement proper error handling and user notifications.


sevvalkaplannn (changes requested):
1-sample static text should appear only in keyword density page, now its in both pages. 
2- since your src/stores/keywordAnalysis.js file has both functionalities for keyword generator and keyword density file its a bit complicated, you can have different stores for each page, if you need one store for common functionalities you can implement it as 3rd.

so you will have
-keywordDensity.js
-keywordGenerator.js
-keyword.js

src/pages/keyword-density/KeywordDensity.vue
```diff
@@ -13,194 +13,73 @@ const props = defineProps({
...
-    
-    MaNotification.info({
-        "size": "large",
-        "variant": "light",
-        "title": "Copied to Clipboard!",
-        "description": "Analysis results have been copied to your clipboard in spreadsheet format.",
-        "type": "info"
-    });
-};
+const store = useKeywordAnalysisStore();
```
sevvalkaplannn, line 16: after you separate the stores please name it like 
const keywordDensityStore = useKeywordDensityStore();
const keywordStore = useKeywordStore();
const keywordGenaratorStore = useKeywordGeneratorStore();

sevvalkaplannn (approved):
very good job ufuk congrats, you can move on to ONB-206

PR #79, opened 2025-07-22
feat(ONB-206): Implement Cypress configuration and end-to-end tests.

ufukagaya:
- Implemented Cypress E2E testing framework with 4 comprehensive test suites.
- Fixed merge keywords notification to show both total keywords and grouped count

- How to run tests:
  yarn dev
  yarn cypress:open

### DIFF
--- a/cypress.config.js
+++ b/cypress.config.js
@@ -1,0 +1,24 @@
+import { defineConfig } from 'cypress'
+
+export default defineConfig({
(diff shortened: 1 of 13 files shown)


ufukagaya:
I realized that the problem was the port configurations left over from the tests I did in the previous version. I re-fixed it to the standard 5173 port.

PR #91, opened 2025-08-14
feat(ONB-204): add keyword counter page with frequency analysis

nagmex210:
- Create VKeywordCounter page with side-by-side responsive layout
  - Implement real-time word count display as user types
  - Add keyword frequency analysis with count and percentage calculation
  - Include responsive TailwindCSS table for results display
  - Add chart-bar icon to card header and navigation tab
  - Support static text prop for initial load from parent pages

### DIFF
--- a/src/pages/VKeywordCounter.vue
+++ b/src/pages/VKeywordCounter.vue
@@ -11,10 +11,32 @@ const props = withDefaults(defineProps<Props>(), {
 });
 
 const inputText = ref('');
-const keywordStats = ref<{ keyword: string; count: number; percentage: number }[]>([]);
-const totalWords = ref(0);
 
-const processText = (text: string) => {
+const totalWords = computed(() => {
+  return textToWords(inputText.value).length;
+});
+
+const wordCount = computed(() => {
+  const words = textToWords(inputText.value);
+  return words.reduce((acc, word) => {
+    acc[word] = (acc[word] || 0) + 1;
+    return acc;
+  }, {} as { [key: string]: number });
+});
+
+const keywordStats = computed(() => {
+  if (totalWords.value === 0) return [];
+  
+  return Object.entries(wordCount.value)
+    .map(([keyword, count]) => ({
+      keyword,
+      count,
+      percentage: Math.round((count / totalWords.value) * 100 * 100) / 100
+    }))
+    .sort((a, b) => b.count - a.count);
+});
+
+const textToWords = (text: string) => {
   if (!text.trim()) return [];
   
   return text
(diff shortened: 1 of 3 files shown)


src/pages/VKeywordCounter.vue
```diff
@@ -0,0 +1,274 @@
...
+
+const props = withDefaults(defineProps<Props>(), {
+  text: ''
+});
+
+const inputText = ref('');
+const keywordStats = ref<{ keyword: string; count: number; percentage: number }[]>([]);
+const totalWords = ref(0);
+
+const processText = (text: string) => {
```
turanibrahim: When I read your code i should understand what is the intend by reading function name. It should be specific for that, It could be long tho. So, we should change function name with better alternative like `textToWords` kinda

src/pages/VKeywordCounter.vue
```diff
@@ -0,0 +1,274 @@
...
+};
+
+onMounted(() => {
+  if (props.text) {
+    inputText.value = props.text;
+    updateWordCount();
+  }
+});
+
+watch(inputText, updateWordCount);
```
turanibrahim: Please don't use watcher it's a not good practice

src/pages/VKeywordCounter.vue
```diff
@@ -0,0 +1,274 @@
...
+const updateWordCount = () => {
+  const words = processText(inputText.value);
+  totalWords.value = words.length;
+};
+
+onMounted(() => {
+  if (props.text) {
+    inputText.value = props.text;
+    updateWordCount();
+  }
```
turanibrahim: I don't recommend inline logic inside life cycle events. That increase complexity in the future. Please move this logic inside a function instead.

src/pages/VKeywordCounter.vue
```diff
@@ -0,0 +1,274 @@
...
+
+watch(inputText, updateWordCount);
+
+const countWords = (words: string[]) => {
+  const wordCount: { [key: string]: number } = {};
+  words.forEach(word => {
+    wordCount[word] = (wordCount[word] || 0) + 1;
+  });
+  return wordCount;
+};
```
turanibrahim: This could be computed we don't need to do this manually

src/pages/VKeywordCounter.vue
```diff
@@ -0,0 +1,274 @@
...
+
+const calculatePercentage = (count: number, total: number) => {
+  return Math.round((count / total) * 100 * 100) / 100;
+};
```
turanibrahim: This also can be computed property

src/pages/VKeywordCounter.vue
```diff
@@ -0,0 +1,274 @@
...
+
+const formatKeywordStats = (wordCount: { [key: string]: number }, totalWords: number) => {
+  return Object.entries(wordCount)
+    .map(([keyword, count]) => ({
+      keyword,
+      count,
+      percentage: calculatePercentage(count, totalWords)
+    }))
+    .sort((a, b) => b.count - a.count);
+};
```
turanibrahim: This one also could be computed property

src/pages/VKeywordCounter.vue
```diff
@@ -0,0 +1,274 @@
...
+              hintText="Paste or type the text you want to analyze for keyword frequency"
+              class="mb-4"
+            />
+            
+            <ma-button 
+              type="primary" 
+              size="medium"
+              @click="analyzeKeywords"
+              class="w-full sm:w-auto"
+            >
```
turanibrahim: Events should be last
```suggestion
            <ma-button 
              type="primary" 
              size="medium"
              class="w-full sm:w-auto"
              @click="analyzeKeywords"
            >
```

turanibrahim (changes requested):

src/pages/VKeywordCounter.vue
```diff
@@ -0,0 +1,248 @@
...
+    .filter(word => word.length > 0);
+};
+
+const initializeText = () => {
+  if (props.text) {
+    inputText.value = props.text;
+  }
+};
+
+onMounted(initializeText);
```
turanibrahim, line 57:
```suggestion
onMounted(() => {
	initializeText()
});
```

When you have multiple operations you can chain them that way

src/pages/VKeywordCounter.vue
```diff
@@ -0,0 +1,248 @@
...
+
+const hasResults = computed(() => keywordStats.value.length > 0);
```
turanibrahim, line 59: haveResults or hasResult

src/pages/VKeywordCounter.vue
```diff
@@ -0,0 +1,248 @@
...
+            
+            <div v-if="inputText.trim()" class="word-count-display">
+              <span class="word-count-label">Total Words:</span>
+              <span class="word-count-value">{{ totalWords }}</span>
+            </div>
+          </div>
+        </ma-card>
+      </div>
+
+      <!-- Results Table Section -->
```
turanibrahim, line 91:
```suggestion
```

src/pages/VKeywordCounter.vue
```diff
@@ -0,0 +1,248 @@
...
+
+onMounted(initializeText);
+
+const hasResults = computed(() => keywordStats.value.length > 0);
+</script>
+
+<template>
+  <div class="keyword-counter-wrapper">
+    <div class="main-container">
+      <!-- Input Section -->
```
turanibrahim, line 65:
```suggestion
```

turanibrahim (changes requested):
We are so close to end. Please update the unit test I would like to have good coverage. Can you add a coverage library to see coverage?

PR #103, opened 2026-06-10
feat: KeywordGenerator.vue modified for ONB-201

okayhamamci:
Implemented the Keyword Generator tool to automatically extract, clean, and display unique 1-gram, 2-gram, and 3-gram keywords from user text input. Duplicates are removed through using "set".

### DIFF
--- a/src/pages/index.vue
+++ b/src/pages/index.vue
@@ -18,23 +18,23 @@
     </div>
 </template>
 
-<style scoped>
+<style lang="scss" scoped>
 .ma-home-container {
     text-align: center;
-    padding-top: 50px;
-}
+    padding-top: 48px;
 
-.ma-keyword-generator-section {
-    margin-top: 40px;
-}
+    .ma-keyword-generator-section {
+        margin-top: 40px;
 
-.ma-keyword-generator-button {
-    display: inline-block;
-    padding: 10px 20px;
-    background-color: #409eff;
-    color: white;
-    text-decoration: none;
-    border-radius: 4px;
-    font-weight: bold;
+        .ma-keyword-generator-button {
+            display: inline-block;
+            padding: 8px 20px;
+            background-color: #409eff;
+            color: white;
+            text-decoration: none;
+            border-radius: 4px;
+            font-weight: bold;
+        }
+    }
 }
 </style>
--- a/src/pages/index.vue
+++ b/src/pages/index.vue
@@ -2,15 +2,15 @@
 
 </script>
 <template>
-    <div class="home-container">
+    <div class="ma-home-container">
         <h1>Welcome to Onboarding Progress</h1>
         <p>You can see the your first task details via link. Please continue that steps while coding.</p>
         <a href="https://mobileaction.atlassian.net/l/cp/7fw8MroM" target="_blank">Implement First Front-End Task</a>
         
-        <div class="keyword-generator-section">
+        <div class="ma-keyword-generator-section">
             <router-link 
                 to="/keyword-generator" 
-                class="keyword-generator-button"
+                class="ma-keyword-generator-button"
             >
                 Open Keyword Generator
             </router-link>
@@ -19,16 +19,16 @@
 </template>
 
 <style scoped>
-.home-container {
+.ma-home-container {
     text-align: center;
     padding-top: 50px;
 }
 
-.keyword-generator-section {
+.ma-keyword-generator-section {
     margin-top: 40px;
 }
 
-.keyword-generator-button {
+.ma-keyword-generator-button {
     display: inline-block;
     padding: 10px 20px;
     background-color: #409eff;
--- a/src/pages/index.vue
+++ b/src/pages/index.vue
@@ -2,18 +2,39 @@
 
 </script>
 <template>
-    <div style="text-align: center; padding-top: 50px;">
+    <div class="home-container">
         <h1>Welcome to Onboarding Progress</h1>
         <p>You can see the your first task details via link. Please continue that steps while coding.</p>
         <a href="https://mobileaction.atlassian.net/l/cp/7fw8MroM" target="_blank">Implement First Front-End Task</a>
         
-        <div style="margin-top: 40px;">
+        <div class="keyword-generator-section">
             <router-link 
                 to="/keyword-generator" 
-                style="display: inline-block; padding: 10px 20px; background-color: #409eff; color: white; text-decoration: none; border-radius: 4px; font-weight: bold;"
+                class="keyword-generator-button"
             >
                 Open Keyword Generator
             </router-link>
         </div>
     </div>
 </template>
+
+<style scoped>
+.home-container {
+    text-align: center;
+    padding-top: 50px;
+}
+
+.keyword-generator-section {
+    margin-top: 40px;
+}
+
+.keyword-generator-button {
+    display: inline-block;
+    padding: 10px 20px;
+    background-color: #409eff;
+    color: white;
+    text-decoration: none;
+    border-radius: 4px;
+    font-weight: bold;
+}
+</style>
--- a/src/pages/index.vue
+++ b/src/pages/index.vue
@@ -2,9 +2,18 @@
 
 </script>
 <template>
-    <div style="text-align: center">
-        <h1>Welcome to Onboarding Progress </h1>
-        <p>You can see the your first task details via link. Please continue that steps while coding. </p>
-        <a href="https://mobileaction.atlassian.net/l/cp/7fw8MroM" target="_blank"> Implement First Front-End Task </a>
+    <div style="text-align: center; padding-top: 50px;">
+        <h1>Welcome to Onboarding Progress</h1>
+        <p>You can see the your first task details via link. Please continue that steps while coding.</p>
+        <a href="https://mobileaction.atlassian.net/l/cp/7fw8MroM" target="_blank">Implement First Front-End Task</a>
+        
+        <div style="margin-top: 40px;">
+            <router-link 
+                to="/keyword-generator" 
+                style="display: inline-block; padding: 10px 20px; background-color: #409eff; color: white; text-decoration: none; border-radius: 4px; font-weight: bold;"
+            >
+                Open Keyword Generator
+            </router-link>
+        </div>
     </div>
 </template>
--- a/src/pages/keyword-generator/KeywordGenerator.vue
+++ b/src/pages/keyword-generator/KeywordGenerator.vue
@@ -1,5 +1,6 @@
 <script setup>
 import { ref, computed } from 'vue';
+import { cleanDescription } from '../../utils/CleanDescription';
 
 const inputText = ref('');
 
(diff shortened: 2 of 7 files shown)


src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,115 @@
...
+    const text = inputText.value;
+    if (!text.trim()) return { oneGrams: [], twoGrams: [], threeGrams: [] };
+
+    // Clean the text: lowercase, replace non-letter/number characters with spaces, 
+    // and split by whitespace
+    const cleanedText = text
+        .toLowerCase()
+        .replace(/[^\p{L}\p{N}\s]/gu, ' ')
+        .replace(/\s+/g, ' ')
+        .trim();
```
sevvalkaplannn: you have `src/cleanupResources.js` utility in project can you use that for cleaning the text,

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,115 @@
...
+    
+    if (!cleanedText) return { oneGrams: [], twoGrams: [], threeGrams: [] };
+
+    const words = cleanedText.split(' ');
+
+    const oneGrams = new Set();
+    const twoGrams = new Set();
+    const threeGrams = new Set();
```
sevvalkaplannn: you're creating a Set just to remove duplicates, then immediately converting it back to an array that's an unnecessary extra step. Since your final output is always an array, you can use array instead of set

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,115 @@
...
+
+    return {
+        oneGrams: Array.from(oneGrams),
+        twoGrams: Array.from(twoGrams),
+        threeGrams: Array.from(threeGrams),
+    };
+});
 </script>
+
 <template>
```
sevvalkaplannn, line 19: In this component, you're using both CSS classes and inline styles at the same time. This makes the code harder to maintain and defeats the purpose of using classes for reusable and organized styling.Can you move the styles out of the template and into a dedicated SCSS section instead of using inline style attributes. Please install [sass](https://sass-lang.com/documentation/) and use SCSS with proper nesting under the main .ma-keywords-generator class to keep the styles clean and more readable.

sevvalkaplannn (changes requested):

src/pages/index.vue
```diff
@@ -2,9 +2,18 @@
...
+        <a href="https://mobileaction.atlassian.net/l/cp/7fw8MroM" target="_blank">Implement First Front-End Task</a>
+        
+        <div style="margin-top: 40px;">
+            <router-link 
+                to="/keyword-generator" 
+                style="display: inline-block; padding: 10px 20px; background-color: #409eff; color: white; text-decoration: none; border-radius: 4px; font-weight: bold;"
+            >
+                Open Keyword Generator
+            </router-link>
+        </div>
```
sevvalkaplannn: can you write css classes for you styling

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,215 @@
 <script setup>
+import { ref, computed } from 'vue';
+import { cleanDescription } from '../../utils/CleanDescription';
+
+const inputText = ref('');
+
+const nGrams = computed(() =>
```
sevvalkaplannn: this computed is handling way too much logic in the component itself. You should extract all the n-gram generation into a utility function, like you did with cleanDescription.
Right now you've got the word splitting, the duplicate checking, and the three separate arrays here move it to a new utility file .js file and call it in here.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,215 @@
...
+
+        // 3-gram
+        if (i < words.length - 2) 
+        {
+            const threeGram = `${words[i]} ${words[i+1]} ${words[i+2]}`;
+            if (!threeGrams.includes(threeGram)) {
+                threeGrams.push(threeGram);
+            }
+        }
+    }
```
sevvalkaplannn: This hardcoded structure doesn't scale at all.
If someone asks for 4-grams later, you're copying and pasting code three times over.
Instead, return an array of objects then in the template, just loop once instead of hardcoding three cards.
Makes everything cleaner and easier to expand.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,215 @@
...
+            <!-- 1-grams -->
+            <div class="gram-card">
+                <h2>
+                    <span>1-Gram</span>
+                    <span class="count-badge">{{ nGrams.oneGrams.length }}</span>
+                </h2>
+                <ul>
+                    <li v-for="kw in nGrams.oneGrams" :key="kw">
+                        {{ kw }}
+                    </li>
```
sevvalkaplannn: kw here is not clearly understandable, when the reviewer reads your code they should understand everything in it, so can you use more descriptive names.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,215 @@
...
+        <div class="input-container">
+            <textarea
+                v-model="inputText"
+                placeholder="Enter your text here (e.g., app description)..."
+                rows="8"
+            ></textarea>
+        </div>
+
+        <div v-if="inputText.trim().length > 0" class="results-grid">
+            <!-- 1-grams -->
```
```suggestion
```
we don't need unnecessary comments in our code, but since you will change the logic you're not gonna write each section again and again after above improvement

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,215 @@
...
+
 <template>
     <div class="ma-keywords-generator">
         <div class="ma-header">
-            <span>Keyword Generator</span>
+            <h1>Keyword Generator</h1>
+            <p>Generate 1,2,3-gram keywords from your text without duplicates.</p>
+        </div>
+
+        <div class="input-container">
```
```suggestion
        <div class="ma-input-container">
```
when writing css classes you need to name it better, for example for .input-containerclass you should name it like input-container, you should add ma-.. for the naming convention (ma means mobileaction for writing company based code), and the names of the CSS classes should be compatible with each other and when we see the class name, it should give information about where it is used.

sevvalkaplannn (changes requested):
I have an additional suggestion,

1- we need to write our commit messages according this structure, can you check this [document](https://mobileaction.atlassian.net/wiki/spaces/Engineerin/pages/1456243537/Conventional+Commit+Messages) again, and then can you send your next commits according that structure.
2- can you re-check the class names and function names, they should explain what they're here for, so naming is very important for us

yarn.lock
sevvalkaplannn, line 1: can you run `yarn` in your terminal, the yarn.lock file is not covering all of the packages

src/pages/index.vue
```diff
@@ -2,9 +2,39 @@
...
+.ma-keyword-generator-button {
+    display: inline-block;
+    padding: 10px 20px;
+    background-color: #409eff;
+    color: white;
+    text-decoration: none;
+    border-radius: 4px;
+    font-weight: bold;
+}
+</style>
```
sevvalkaplannn: .ma-keyword-generator-button is a child of .ma-keyword-generator-section and two of them are are direct children of .ma-home-container in the template but in CSS, they're defined at the root level instead of nested inside. The main goal here is to reflect the parent-child relationship in HTML in CSS nesting as well.

```suggestion
.ma-home-container {
  text-align: center;
  padding-top: 50px;

  .ma-keyword-generator-section {
    margin-top: 40px;

    .ma-keyword-generator-button {
      display: inline-block;
      padding: 10px 20px;
      background-color: #409eff;
      color: white;
      text-decoration: none;
      border-radius: 4px;
      font-weight: bold;
    }
  }
}
</style>
```

src/utils/generateNGrams.js
```diff
@@ -0,0 +1,25 @@
+export function generateNGrams(cleanedText, maxN = 3) {
```
sevvalkaplannn:
```suggestion
const MAX_N_GRAM_VALUE = 3
export function generateNGrams(cleanedText, maxN = MAX_N_GRAM_VALUE) {
```
it is better to make the static and constant values as consts in the files for its readability and reusability.

src/pages/index.vue
```diff
@@ -2,9 +2,39 @@
...
+            >
+                Open Keyword Generator
+            </router-link>
+        </div>
     </div>
-</template>
\ No newline at end of file
+</template>
+
+<style scoped>
```
sevvalkaplannn:
```suggestion
<style lang="scss" scoped>
```

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,148 @@
...
+                </ul>
+                <div v-if="nGramCategory.keywords.length === 0" class="ma-empty-results-message">No {{ nGramCategory.id }}-grams generated</div>
+            </div>
         </div>
     </div>
-</template>
\ No newline at end of file
+</template>
+
+<style lang="scss" scoped>
```
sevvalkaplannn, line 51: its just a quick tip, we use multiples of 4 for px sizes, it's cleaner, more maintainable, and aligns with our spacing system where everything is based on 4px units. Also later when we migrate to Tailwind, it'll be much easier, Tailwind has predefined scales (text-sm, text-base, text-lg, etc.) based on multiples of 4, so custom arbitrary values like 14px won't map cleanly. Can you check the whole css's according this rule.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,148 @@
...
+import { ref, computed } from 'vue';
+import { cleanDescription } from '../../utils/CleanDescription';
+import { generateNGrams } from '../../utils/generateNGrams';
+
+const sourceDescriptionText = ref('');
+
+const generatedKeywordNGrams = computed(() => 
+{
+    const text = sourceDescriptionText.value;
+    if (!text.trim()) return { singleWordKeywords: [], twoWordKeywords: [], threeWordKeywords: [] };
```
sevvalkaplannn: you're returning object but below you return array from generateNGrams, template expects array, can you return array

sevvalkaplannn: two word keywords or three word keywords are not explaining our case, since for example 2-gram keywords are not always 2 word keyword, can you rename

src/utils/generateNGrams.js
```diff
@@ -0,0 +1,25 @@
...
+                    keywords.push(nGram);
+                }
+            }
+        }
+
+        results.push({
+            id: n,
+            title: `${n}-Gram`,
+            keywords: keywords
+        });
```
sevvalkaplannn, line 23: always pushing, even empty results check if there are keywords (keywords.length > 0)

src/utils/generateNGrams.js
```diff
@@ -0,0 +1,25 @@
...
+    for (let n = 1; n <= maxN; n++) {
+        const keywords = [];
+
+        if (cleanedWordsArray.length >= n) {
+            for (let i = 0; i <= cleanedWordsArray.length - n; i++) {
+                const nGram = cleanedWordsArray.slice(i, i + n).join(' ');
+                if (nGram && !keywords.includes(nGram)) {
+                    keywords.push(nGram);
+                }
+            }
```
sevvalkaplannn: you're using keywords.includes() to check if a keyword already exists, first create the array, and then check, that will be more performance friendly

sevvalkaplannn (changes requested):
1- we need to write our commit messages according this structure, can you check this [document](https://mobileaction.atlassian.net/wiki/spaces/Engineerin/pages/1456243537/Conventional+Commit+Messages) again, and then can you send your next commits according that structure

for example your last commit should be like
"chore(ONB-201): Naming is modified and yarn.lock is updated"
since we wrote our commit is belongs to ONB-201 you don't need to write "for ONB-201", and chore, fix, refactor, feat keys are explained in the above link. 

Image

src/utils/generateNGrams.js
```diff
@@ -1,19 +1,20 @@
...
-                    keywords.push(nGram);
-                }
-            }
+        const nGrams = [];
+        for (let i = 0; i <= cleanedWordsArray.length - n; i++) {
+            nGrams.push(cleanedWordsArray.slice(i, i + n).join(' '));
         }
 
+        const keywords = [...new Set(nGrams)];
+        if (keywords.length === 0) continue;
```
sevvalkaplannn: we don't need this line we can delete this if check,  i`f (cleanedWordsArray.length < n) continue;` is already handles that case

src/utils/generateNGrams.js
```diff
@@ -1,19 +1,20 @@
...
-                if (nGram && !keywords.includes(nGram)) {
-                    keywords.push(nGram);
-                }
-            }
+        const nGrams = [];
+        for (let i = 0; i <= cleanedWordsArray.length - n; i++) {
+            nGrams.push(cleanedWordsArray.slice(i, i + n).join(' '));
         }
 
+        const keywords = [...new Set(nGrams)];
```
sevvalkaplannn: we don't need that, we can make a removeDuplicates utility on utils folder and use it in here, creating Set from an array is not recommended

sevvalkaplannn (changes requested):

sevvalkaplannn (approved):
thanks for your work, you can push that pr to your main branch and continue with ONB-202 🙌🏻

PR #104, opened 2026-06-18
ONB-201: Add Keyword Generator Page

cmlcz:
Add a new page where you paste in text and get a list of keywords extracted from the text. It works by breaking the text into 1, 2, 3-word phrases (n-grams).


sevvalkaplannn (changes requested):
I have some suggestions before reviewing your code
1- Can you add a button into landing page that redirects us into keyword generator page

https://github.com/user-attachments/assets/c2876fe4-52e9-4056-9ffd-a2aa9024575b

2- We can add more styling to the n-gram results section, as it currently feels quite squeezed. Also, even after deleting all keywords, the n-gram results are still visible. It might be helpful to add a reset button that clears the input and resets the results.

https://github.com/user-attachments/assets/ff205380-797d-44f8-b922-53e641701c48

3-

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,109 @@
...
+            </div>
+            <p v-else-if="error" style="color: red;">{{ error }}</p>
+            <p v-else>Results will be in here</p>
         </div>
     </div>
-</template>
\ No newline at end of file
+</template>
+
+<style scoped>
```
sevvalkaplannn: can you install sass into your project and write scss, sass (scss) is basically CSS with extra features that make styling easier and cleaner, like variables, nesting writing, and reusable code blocks. It gets compiled into regular CSS before being sent to the browser, but its better to write the stylings. After installation can you write those css's in nested way. https://sass-lang.com/documentation/

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,109 @@
...
+.ma-layout {
+    display: flex;
+    max-width: 960px;
+    margin: 48px auto;
+    padding: 0 20px;
+    font-family: system-ui, sans-serif;
+}
+.ma-side {
```
sevvalkaplannn: when writing css classes you need to name it better, and the names of the CSS classes should be compatible with each other and when we see the class name, it should give information about where it is used. According this suggestion can you change the class names.

src/utils/NGramUtils.js
```diff
@@ -0,0 +1,44 @@
...
+    if(textArray.length === 0) return null;
+
+    const result = [];
+
+    for(let i=1;i<=3;i++){
+        result.push({ngram: i, keywords: generateNGram(textArray, i)});
+    }
+
+    return result;
+}
```
sevvalkaplannn: its a utility file, we should be able to use it in anywhere we want for this functionality but now it only handles 1 to 3 grams, make it configurable instead of hardcoding 1 to 3

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,109 @@
...
+        <div class="ma-main">
+            <div v-if="keywords">
+                <div v-for="g in keywords" :key="g.ngram">
+                    <strong>{{ g.ngram }}-gram ({{ g.keywords.length }})</strong>
+                    <div>
+                        <span v-for="kw in g.keywords" :key="kw">{{ kw }}</span>
+                    </div>
+                </div>
+            </div>
+            <p v-else-if="error" style="color: red;">{{ error }}</p>
```
```suggestion
            <p v-else-if="error">{{ error }}</p>
```
we don't need inline styles, you can use tailwind for inline class in later tasks, but since we are only writing css now, you should make all the styling as classes, it works like your way too, but its not readable since you have both classes and inline styling.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,109 @@
...
+            <textarea v-model="inputText" rows="8"></textarea>
+            <button @click="generate">Generate</button>
+            <button @click="resetKeywords">Reset Keywords</button>
+        </div>
+        <div class="ma-main">
+            <div v-if="keywords">
+                <div v-for="g in keywords" :key="g.ngram">
+                    <strong>{{ g.ngram }}-gram ({{ g.keywords.length }})</strong>
+                    <div>
+                        <span v-for="kw in g.keywords" :key="kw">{{ kw }}</span>
```
sevvalkaplannn: g here is difficult to understand, when the reviewer reads your code they should understand everything in it, so can you use more descriptive names.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,109 @@
...
+function generate() {
+    error.value = '';
+    keywords.value = generateKeyword(inputText.value);
+
+    if(!keywords.value){
+        error.value = "Enter text!";
+        keywords.value = null;
+        return;
+    }
+}
```
sevvalkaplannn, line 24: the validation check happens after the function call instead of before, so the error message appears after unnecessary processing is complete, this pattern is backwards you should validate the input first, and only call generateKeyword() if validation passes, preventing wasted computation entirely.

src/pages/index.vue
```diff
@@ -1,10 +1,31 @@
...
+        <a href="https://mobileaction.atlassian.net/l/cp/7fw8MroM" target="_blank" class="ma-routing" style="text-decoration: none;"> Implement First Front-End Task </a>
+        <button @click="goToKeywordPage" class="ma-routing">Keyword Generator</button>
     </div>
-</template>
\ No newline at end of file
+</template>
+
+<style scoped>
+.ma-routing{
+    margin: 10px 10px;
```
sevvalkaplannn:
```suggestion
    margin: 10px;
```

    margin: 10px 10px; means top/bottom 10px, left/right 10px

you can directly write margin: 10px; since every direction has same px size.

src/pages/index.vue
```diff
@@ -1,10 +1,31 @@
...
+    padding: 8px 24px;
+    border: none;
+    border-radius: 6px;
+    background: #1a1a1a;
+    color: #fff;
+    font-size: 15px;
```
sevvalkaplannn: its just a quick tip, we use multiples of 4 for font sizes, it's cleaner, more maintainable, and aligns with our spacing system where everything is based on 4px units. Also later when we migrate to Tailwind, it'll be much easier, Tailwind has predefined scales (text-sm, text-base, text-lg, etc.) based on multiples of 4, so custom arbitrary values like 15px won't map cleanly.

sevvalkaplannn: can you check the other font-sizes too

sevvalkaplannn (changes requested):
I have two additional suggestions for your pr cemil, 

1- we need to write our commit messages according this structure, can you check this [file](https://mobileaction.atlassian.net/wiki/spaces/Engineerin/pages/1456243537/Conventional+Commit+Messages) again, and then can you send your next commits according that structure.

2- Could you write a description in the github PR section in english that everyone can understand

src/pages/keyword-generator/KeywordGenerator.vue
cmlcz: For example, if input are "?!.*" then the input is not empty but not valuable because we can't extract any word from it. So, we should also check it after the function call. But I'll add another check before function call because it can be empty("") and we don't need to continue.

src/pages/index.vue
```diff
@@ -1,10 +1,32 @@
...
         <p>You can see the your first task details via link. Please continue that steps while coding. </p>
-        <a href="https://mobileaction.atlassian.net/l/cp/7fw8MroM" target="_blank"> Implement First Front-End Task </a>
+        <a href="https://mobileaction.atlassian.net/l/cp/7fw8MroM" target="_blank" class="ma-routing"> Implement First Front-End Task </a>
+        <button @click="goToKeywordPage" class="ma-routing">Keyword Generator</button>
     </div>
-</template>
\ No newline at end of file
+</template>
+
+<style scoped>
```
sevvalkaplannn:
```suggestion
<style lang="scss" scoped>
```

src/utils/NGramUtils.js
```diff
@@ -0,0 +1,32 @@
+import { filterArr } from "@/cleanupResources";
```
sevvalkaplannn: 'filterArr' is declared but its value is never read.ts(6133)

file has an error on this line, you're importing filterArr but never using it.

src/utils/NGramUtils.js
```diff
@@ -0,0 +1,32 @@
...
+import { cleanDescription } from "./CleanDescription";
+
+export const generateKeyword = (text, min = 1, max = 3) => {
```
sevvalkaplannn: person who reads the file couldn't understand these edge numbers (1 and 3), you can create constants for those static numbers for example 

const MIN_NGRAM = 1
const MAX_NGRAM = 3

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,134 @@
...
+            </div>
+            <p v-else-if="error" class="ma-error-text">{{ error }}</p>
+            <p v-else>Results will be in here</p>
         </div>
     </div>
-</template>
\ No newline at end of file
+</template>
+
+<style lang="scss" scoped>
```
sevvalkaplannn, line 56: .ma-input-panel and .ma-results-panel are direct children of .ma-keyword-generator in the template but in CSS, they're defined at the root level instead of nested inside. The main goal here is to reflect the parent-child relationship in HTML in CSS nesting as well.

src/utils/NGramUtils.js
```diff
@@ -0,0 +1,32 @@
...
+        const keywords = generateNGram(textArray, i);
+        if(keywords.length > 0) result.push({ngram: i, keywords});
+    }
+
+    return result;
+}
+
+const generateNGram = (array, ngram) => {
+
+    const newArray = [];
```
sevvalkaplannn: newArray is not a good name. Call it nGrams or generatedNGrams, the namings are very important for us

src/utils/NGramUtils.js
```diff
@@ -0,0 +1,32 @@
...
+
+    for(let i = 0; i <= array.length - ngram; i++){
+        const element = array.slice(i, i + ngram).join(' ');
+        newArray.push(element);
+    }
+
+    return [...new Set(newArray)];
```
sevvalkaplannn: Instead of set usage, use the removeDuplicates() function pattern you already have. This keeps your deduplication logic consistent across the codebase instead of mixing different patterns.

sevvalkaplannn (changes requested):
You can resolve my comments after you are done with them

package.json
sevvalkaplannn, line 1: can you run `yarn` in your terminal, the yarn.lock file is not created

src/utils/NGramUtils.js
```diff
@@ -21,12 +23,27 @@ export const generateKeyword = (text, min = 1, max = 3) => {
...
+
+    for (const item of array){
+        if(!seen.has(item)){
+            seen.add(item);
+            result.push(item);
+        }
+    }
+
+    return result;
 }
```
sevvalkaplannn, line 47: you don't need to create a set for seen, you can continue with same array

sevvalkaplannn (approved):
thanks for your work, you can push that pr to your main branch and continue with ONB-202 🙌🏻

PR #105, opened 2026-06-22
ONB-202: add action-kit UI and selectable n-gram sizes to keyword generator

cmlcz:
This updates the Keyword Generator page to use our Action-Kit UI components instead of plain HTML, CSS. It also adds a multi-select options where you can pick which n-gram sizes you want (1 to 10), instead of getting fixed 1 to 3. The keyword logic now takes those picked sizes and skips common stopwords.


sevvalkaplannn (changes requested):
you need to write the css's with tailwind classes

src/pages/index.vue
sevvalkaplannn, line 1: can you refactor this file too, you can use MaButton in here and you can align the landing texts. Since this is your main page in your project, it should have better ui.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -15,120 +23,47 @@ function generate() {
...
 </script>
 
 <template>
-    <div class="ma-keyword-generator">
-        <div class="ma-input-panel">
-            <h2>Keyword Generator</h2>
-            <textarea v-model="inputText" rows="8"></textarea>
-            <button @click="generate">Generate</button>
-            <button @click="resetKeywords">Reset Keywords</button>
+    <div class="flex gap-10 max-w-[960px] my-12 mx-auto px-5 font-sans text-[#2c3e50]">
```
sevvalkaplannn: tailwind's purpose is obtaining consistency through design tokens but in this file almost every color and size here is written as a custom value, can you refactor them with tailwind classes for example for max-w-[960px] you can use max-w-240 or for text-[#2c3e50] you can use text-gray-800

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -15,120 +23,47 @@ function generate() {
...
+        <div class="flex flex-col gap-3 basis-[400px] shrink-0">
```
```suggestion
        <div class="flex flex-col gap-3">
```
especially basis usage is not necessary for our project, can you re-check the css's and simplify them

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -15,120 +23,47 @@ function generate() {
...
+            <h2 class="m-0">Keyword Generator</h2>
```
```suggestion
            <h2>Keyword Generator</h2>
```
do you need that

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -15,120 +23,47 @@ function generate() {
...
+                <div v-for="group in keywords" :key="group.ngram" class="p-4 bg-[#fafafa] rounded-lg mb-5">
+                    <strong class="block mb-3 text-base">{{ group.ngram }}-gram ({{ group.keywords.length }})</strong>
+                    <div class="flex flex-wrap gap-2">
+                        <MaBadge v-for="keyword in group.keywords" :key="keyword" size="large" type="secondary" variant="teal">{{ keyword }}</MaBadge>
                     </div>
                 </div>
             </div>
-            <p v-else-if="error" class="ma-error-text">{{ error }}</p>
-            <p v-else>Results will be in here</p>
+            <p v-else-if="error" class="text-[red]">{{ error }}</p>
```
```suggestion
            <p v-else-if="error" class="text-red-500">{{ error }}</p>
```
it should work

sevvalkaplannn (changes requested):
can you check the tailwind classes again, there are some wrong or redundant usages.

sevvalkaplannn (changes requested):
1- you can show the error on MaSelect itself, you can use hasError and hint prop.

Image

Image

2- you can show the error on MaTextarea with error prop

3- Adding button for submitting the keyword list is on 3rd task, in this task we need to show the results dynamically without user's submission. But you can keep the reset button. (You can comment the existing button implementation for next task.)

4- Can you resolve the other comments on pr. It is important for us to maintain the review process.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,12 +1,20 @@
...
 
-function generate() {
+const nGramSizes = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+const nGramOptions = nGramSizes.map((size) => ({
+    label: `${size}-gram`,
+    value: size,
+}));
+const selectedNGrams = ref([]);
+
+function generate(){
```
sevvalkaplannn: consider using generateKeywords instead of generate to make the function name more descriptive and consistent with resetKeywords

sevvalkaplannn (changes requested):

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,12 +1,20 @@
 <script setup>
 import { ref } from 'vue';
 import { generateKeyword } from '@/utils/NGramUtils';
+import { MaBadge, MaTextarea, MaButton, MaSelect2 } from '@mobileaction/action-kit';
```
sevvalkaplannn: Its not mandatory but you can use imports like below, 
```suggestion
import { MaBadge, MaTextarea, MaButton, MaSelect2 as MaSelect } from '@mobileaction/action-kit';
```

and then use <MaSelect /> in the template.

this makes the component name shorter and easier to read while still preserving the original import source. It also keeps the template cleaner, especially when the component is used multiple times.

sevvalkaplannn (approved):
1- this commit is not a feat commit, it is chore, can you check the [document](https://mobileaction.atlassian.net/wiki/spaces/Engineerin/pages/1456243537/Conventional+Commit+Messages) for your next commits.

thanks for the work you can continue with ONB-203

PR #106, opened 2026-06-23
feat(ONB-202): Stop word functionality is added.

okayhamamci:
ActionKit is added. Possible n value increased to 10 with multi-select option.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,10 @@
     "lint": "eslint . --ext .vue,.js,.jsx,.cjs,.mjs --fix --ignore-path .gitignore"
   },
   "dependencies": {
-    "vue": "^3.4.29",
+    "@mobileaction/action-kit": "^1.58.18",
+    "@tailwindcss/vite": "^4.3.1",
+    "tailwindcss": "^4.3.1",
+    "vue": "^3.5.38",
     "vue-router": "^4.3.3"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "@mobileaction/action-kit": "^1.58.18",
     "@tailwindcss/vite": "^4.3.1",
     "tailwindcss": "^4.3.1",
-    "vue": "^3.5.38",
+    "vue": "^3.4.29",
     "vue-router": "^4.3.3"
   },
   "devDependencies": {
--- a/src/components/TheNav.vue
+++ b/src/components/TheNav.vue
@@ -1,16 +1,7 @@
 <template>
-    <div class="ma-router-item">
+    <div class="flex flex-row justify-evenly m-[15px]">
         <router-link to="/keyword-density">
             Keyword Density
         </router-link>
     </div>
 </template>
-
-<style>
- .ma-router-item {
-   display: flex;
-   flex-direction: row;
-   justify-content: space-evenly;
-   margin: 15px;
- }
-</style>
--- a/src/components/TheNav.vue
+++ b/src/components/TheNav.vue
@@ -1,7 +1,16 @@
 <template>
-    <div class="flex flex-row justify-evenly m-[15px]">
+    <div class="ma-router-item">
         <router-link to="/keyword-density">
             Keyword Density
         </router-link>
     </div>
 </template>
+
+<style>
+ .ma-router-item {
+   display: flex;
+   flex-direction: row;
+   justify-content: space-evenly;
+   margin: 15px;
+ }
+</style>
--- a/src/pages/index.vue
+++ b/src/pages/index.vue
@@ -10,7 +10,7 @@
         <div class="mt-10">
             <router-link 
                 to="/keyword-generator" 
-                class="inline-block px-5 py-2 bg-[#409eff] text-white no-underline rounded font-bold"
+                class="inline-block px-5 py-2 bg-blue-500 text-white no-underline rounded font-bold"
             >
                 Open Keyword Generator
             </router-link>
--- a/src/pages/index.vue
+++ b/src/pages/index.vue
@@ -1,19 +1,26 @@
 <script setup>
+import { MaButton } from '@mobileaction/action-kit';
+import { useRouter } from 'vue-router';
 
+const router = useRouter();
+
+const openTaskLink = () => {
+    window.open('https://mobileaction.atlassian.net/l/cp/7fw8MroM', '_blank');
+};
 </script>
 <template>
-    <div class="text-center pt-12">
-        <h1>Welcome to Onboarding Progress</h1>
-        <p>You can see the your first task details via link. Please continue that steps while coding.</p>
-        <a href="https://mobileaction.atlassian.net/l/cp/7fw8MroM" target="_blank">Implement First Front-End Task</a>
+    <div class="flex flex-col items-center justify-center pt-12 space-y-6 max-w-2xl mx-auto text-center font-sans">
+        <h1 class="text-3xl font-bold text-slate-800 m-0">Welcome to Onboarding Progress</h1>
+        <p class="text-gray-500 text-lg">You can see the your first task details via link. Please continue that steps while coding.</p>
         
-        <div class="mt-10">
-            <router-link 
-                to="/keyword-generator" 
-                class="inline-block px-5 py-2 bg-blue-500 text-white no-underline rounded font-bold"
-            >
+        <div class="flex flex-col sm:flex-row items-center gap-4 mt-6">
+            <MaButton @click="openTaskLink">
+                Implement First Front-End Task
+            </MaButton>
+            
+            <MaButton type="primary" @click="router.push('/keyword-generator')">
                 Open Keyword Generator
-            </router-link>
+            </MaButton>
         </div>
     </div>
 </template>
--- a/src/pages/index.vue
+++ b/src/pages/index.vue
@@ -2,39 +2,18 @@
 
 </script>
 <template>
-    <div class="ma-home-container">
+    <div class="text-center pt-12">
         <h1>Welcome to Onboarding Progress</h1>
         <p>You can see the your first task details via link. Please continue that steps while coding.</p>
         <a href="https://mobileaction.atlassian.net/l/cp/7fw8MroM" target="_blank">Implement First Front-End Task</a>
         
-        <div class="ma-keyword-generator-section">
+        <div class="mt-10">
             <router-link 
                 to="/keyword-generator" 
-                class="ma-keyword-generator-button"
+                class="inline-block px-5 py-2 bg-[#409eff] text-white no-underline rounded font-bold"
             >
                 Open Keyword Generator
             </router-link>
         </div>
     </div>
 </template>
-
-<style lang="scss" scoped>
-.ma-home-container {
-    text-align: center;
-    padding-top: 48px;
-
-    .ma-keyword-generator-section {
-        margin-top: 40px;
-
-        .ma-keyword-generator-button {
-            display: inline-block;
-            padding: 8px 20px;
-            background-color: #409eff;
-            color: white;
-            text-decoration: none;
-            border-radius: 4px;
-            font-weight: bold;
-        }
-    }
-}
-</style>
--- a/src/pages/keyword-generator/KeywordGenerator.vue
+++ b/src/pages/keyword-generator/KeywordGenerator.vue
@@ -32,152 +32,53 @@ const generatedKeywordNGrams = computed(() => {
 </script>
 
 <template>
-    <div class="ma-keyword-generator-page-wrapper">
-        <div class="ma-page-header-section">
-            <h1>Keyword Generator</h1>
-            <p>Generate 1-10 gram keywords from your text without duplicates.</p>
+    <div class="p-5 max-w-[1000px] mx-auto font-sans">
+        <div class="mb-6">
+            <h1 class="text-[28px] font-bold mb-2 text-[#2c3e50]">Keyword Generator</h1>
+            <p class="text-[#666] text-base m-0">Generate 1-10 gram keywords from your text without duplicates.</p>
         </div>
 
-        <div class="ma-filters-section">
+        <div class="mb-6 flex items-center justify-between">
             <MaSelect2
                 v-model:value="selectedNGrams"
                 :options="nGramOptions"
                 multiple
                 placeholder="Select N-Grams to generate"
-                class="ma-ngram-selector"
+                class="w-full max-w-[400px]"
             />
             <MaCheckbox v-model:checked="shouldRemoveStopWords">
                 Remove Stop Words
             </MaCheckbox>
         </div>
 
-        <div class="ma-text-input-section">
+        <div class="mb-8">
             <MaInput2
                 v-model="sourceDescriptionText"
                 placeholder="Enter your text here (e.g., app description)..."
                 :rows="8"
             />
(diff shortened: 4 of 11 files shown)


package.json
```diff
@@ -10,7 +10,10 @@
     "lint": "eslint . --ext .vue,.js,.jsx,.cjs,.mjs --fix --ignore-path .gitignore"
   },
   "dependencies": {
-    "vue": "^3.4.29",
+    "@mobileaction/action-kit": "^1.58.18",
+    "@tailwindcss/vite": "^4.3.1",
+    "tailwindcss": "^4.3.1",
+    "vue": "^3.5.38",
```
sevvalkaplannn: is there a reason that you updated vue's version

sevvalkaplannn (changes requested):
you need to change css's with the tailwind classes

package.json
okayhamamci: I thought there might be a problem related to version mismatch during installation of tailwindcss, so I updated it. There is not any specific reason other than that.

src/components/TheNav.vue
sevvalkaplannn, line 1: this navbar is not imported in your project right now, and keyword density is not implemented yet, you can implement the navbar on ONB-204, right now we don't need it.

src/pages/index.vue
sevvalkaplannn, line 1: you can use MaButton in here for both buttons (first project button and go tokeyword generator button) and you can align the landing texts. Since this is your main page in your project, it should have better ui.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -2,147 +2,83 @@
...
+            <div class="mt-3 flex justify-end" v-if="sourceDescriptionText.trim().length > 0">
+                <MaButton @click="sourceDescriptionText = ''">Clear Text</MaButton>
+            </div>
         </div>
 
-        <div v-if="sourceDescriptionText.trim().length > 0" class="ma-ngram-results-grid">
-            <div v-for="nGramCategory in generatedKeywordNGrams" :key="nGramCategory.id" class="ma-ngram-category-card">
-                <h2>
+        <div v-if="sourceDescriptionText.trim().length > 0" class="grid gap-6 grid-cols-[repeat(auto-fit,minmax(248px,1fr))]">
+            <div v-for="nGramCategory in generatedKeywordNGrams" :key="nGramCategory.id" class="bg-[#f8f9fa] p-5 rounded-xl border border-[#ebeef5] shadow-[0_2px_12px_0_rgba(0,0,0,0.05)]">
```
sevvalkaplannn: tailwind's purpose is obtaining consistency through design, but in this file almost every color here is written as a custom value, can you refactor them with tailwind classes for example use border-gray-200 instead of border-[#ebeef5], for bg-[#f8f9fa] likewise, for custom shadow classes you can use existing ones.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -2,147 +2,83 @@
...
+            />
+            <div class="mt-3 flex justify-end" v-if="sourceDescriptionText.trim().length > 0">
+                <MaButton @click="sourceDescriptionText = ''">Clear Text</MaButton>
+            </div>
         </div>
 
-        <div v-if="sourceDescriptionText.trim().length > 0" class="ma-ngram-results-grid">
-            <div v-for="nGramCategory in generatedKeywordNGrams" :key="nGramCategory.id" class="ma-ngram-category-card">
-                <h2>
+        <div v-if="sourceDescriptionText.trim().length > 0" class="grid gap-6 grid-cols-[repeat(auto-fit,minmax(248px,1fr))]">
```
sevvalkaplannn: grid is a good option, but you can use flex too. just a note. https://tailwindcss.com/docs/flex

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -2,147 +2,83 @@
...
+            <div v-for="nGramCategory in generatedKeywordNGrams" :key="nGramCategory.id" class="bg-[#f8f9fa] p-5 rounded-xl border border-[#ebeef5] shadow-[0_2px_12px_0_rgba(0,0,0,0.05)]">
+                <h2 class="text-[20px] mb-4 text-[#303133] flex items-center justify-between mt-0">
```
sevvalkaplannn: custom values are not the best approach when writing tailwind, and text-[20px] has exact class on tailwind text-xl is font-size: 20px.

```suggestion
                <h2 class="text-xl mb-4 text-gray-900 flex items-center justify-between mt-0">
```

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -2,147 +2,83 @@
...
+    { value: 2, label: '2-Gram' },
+    { value: 3, label: '3-Gram' },
+    { value: 4, label: '4-Gram' },
+    { value: 5, label: '5-Gram' },
+    { value: 6, label: '6-Gram' },
+    { value: 7, label: '7-Gram' },
+    { value: 8, label: '8-Gram' },
+    { value: 9, label: '9-Gram' },
+    { value: 10, label: '10-Gram' },
+];
```
sevvalkaplannn: nGramOptions array is repetitive and could be simplified

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -2,147 +2,83 @@
...
+
+const generatedKeywordNGrams = computed(() => {
```
sevvalkaplannn: if users can select values in any order, the generated sections may appear in an unexpected order. Consider sorting selectedNGrams before generating results

Image

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -2,147 +2,83 @@
...
+        <div class="mb-8">
+            <MaInput2
                 v-model="sourceDescriptionText"
                 placeholder="Enter your text here (e.g., app description)..."
-                rows="8"
-            ></textarea>
+                :rows="8"
+            />
+            <div class="mt-3 flex justify-end" v-if="sourceDescriptionText.trim().length > 0">
+                <MaButton @click="sourceDescriptionText = ''">Clear Text</MaButton>
```
```suggestion
                <MaButton @click="resetKeywordInput>Clear Text</MaButton>
```
we should make the function on script part.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -2,147 +2,83 @@
...
 
-        <div v-if="sourceDescriptionText.trim().length > 0" class="ma-ngram-results-grid">
-            <div v-for="nGramCategory in generatedKeywordNGrams" :key="nGramCategory.id" class="ma-ngram-category-card">
-                <h2>
+        <div v-if="sourceDescriptionText.trim().length > 0" class="grid gap-6 grid-cols-[repeat(auto-fit,minmax(248px,1fr))]">
+            <div v-for="nGramCategory in generatedKeywordNGrams" :key="nGramCategory.id" class="bg-[#f8f9fa] p-5 rounded-xl border border-[#ebeef5] shadow-[0_2px_12px_0_rgba(0,0,0,0.05)]">
+                <h2 class="text-[20px] mb-4 text-[#303133] flex items-center justify-between mt-0">
                     <span>{{ nGramCategory.title }}</span>
-                    <span class="ma-keyword-count-badge">{{ nGramCategory.keywords.length }}</span>
+                    <span class="text-base bg-[#e0e6ed] py-1 px-2 rounded-xl text-[#606266]">{{ nGramCategory.keywords.length }}</span>
```
sevvalkaplannn: you can use MaBadge too.

src/pages/keyword-generator/KeywordGenerator.vue
sevvalkaplannn: you can use MaCard instead of creating one since we already have it

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -2,147 +2,83 @@
...
                 </h2>
-                <ul>
-                    <li v-for="keyword in nGramCategory.keywords" :key="keyword">
+                <div class="flex flex-wrap gap-2 max-h-[400px] overflow-y-auto">
+                    <MaBadge 
+                        v-for="keyword in nGramCategory.keywords" 
+                        :key="keyword"
+                        class="mr-1 mb-1"
```
sevvalkaplannn: we should avoid overriding their appearance or relying on custom colors and spacing unless there is a strong product requirement
```suggestion
```

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -2,147 +2,83 @@
 import { ref, computed } from 'vue';
 import { cleanDescription } from '../../utils/CleanDescription';
 import { generateNGrams } from '../../utils/generateNGrams';
+import { MaTextarea as MaInput2, MaSelect2, MaBadge, MaButton, MaCheckbox } from '@mobileaction/action-kit';
```
```suggestion
import { MaTextarea, MaSelect2 as MaSelect, MaBadge, MaButton, MaCheckbox } from '@mobileaction/action-kit';
```
MaTextArea and MaInput2 are not the same components so its a wrong casting, that kind of castings can be done with MaSelect2 cases

sevvalkaplannn: clarifying about the actionkit components if you take a look at the storybook you will see some components have v1 and v2, for example MaCheckbox is deprecated, instead of that you use MaCheckbox2 like you did in the MaSelect2 case

sevvalkaplannn (changes requested):
1- you can use MaCard for your n-gram card's
2- can you check the whole custom tailwind classes.

the page currently looks more like a standalone custom ui than a screen built with actionkit components. We should try to lean on the design system for both styling and layout before introducing custom ui patterns since we have design system.

src/pages/keyword-generator/KeywordGenerator.vue
sevvalkaplannn: this makes the component name shorter and easier to read while still preserving the original import source. It also keeps the template cleaner, especially when the component is used multiple times.

src/pages/index.vue
```diff
@@ -2,39 +2,18 @@
...
         
-        <div class="ma-keyword-generator-section">
+        <div class="mt-10">
             <router-link 
                 to="/keyword-generator" 
-                class="ma-keyword-generator-button"
+                class="inline-block px-5 py-2 bg-blue-500 text-white no-underline rounded font-bold"
             >
                 Open Keyword Generator
             </router-link>
```
sevvalkaplannn: you can use MaButton in here

src/pages/index.vue
```diff
@@ -2,39 +2,18 @@
 
 </script>
 <template>
-    <div class="ma-home-container">
+    <div class="text-center pt-12">
         <h1>Welcome to Onboarding Progress</h1>
         <p>You can see the your first task details via link. Please continue that steps while coding.</p>
         <a href="https://mobileaction.atlassian.net/l/cp/7fw8MroM" target="_blank">Implement First Front-End Task</a>
```
sevvalkaplannn: you can use MaButton in here

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -2,147 +2,80 @@
...
+                        :key="keyword"
+                    >
                         {{ keyword }}
-                    </li>
-                </ul>
-                <div v-if="nGramCategory.keywords.length === 0" class="ma-empty-results-message">No {{ nGramCategory.id }}-grams generated</div>
-            </div>
+                    </MaBadge>
+                </div>
+                <MaEmpty v-if="nGramCategory.keywords.length === 0" :description="`No ${nGramCategory.id}-grams generated`" />
```
sevvalkaplannn: I'm not able to see this case on ui

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -2,147 +2,80 @@
...
+
+        <div class="mb-8">
+            <MaTextarea
                 v-model="sourceDescriptionText"
                 placeholder="Enter your text here (e.g., app description)..."
-                rows="8"
-            ></textarea>
+                :rows="8"
+            />
+            <div class="mt-3 flex justify-end" v-if="sourceDescriptionText.trim().length > 0">
```
sevvalkaplannn: sourceDescriptionText.trim() is used multiple places on this file this logic should be derived once in script and reused.

sevvalkaplannn: it can be a computed like 
`const hasInput = computed(() => sourceDescriptionText.value.trim().length > 0);`

and use it as `v-if="hasInput"` on template, it is better for, readability, single source of truth and with that we avoid repeated string processing during render

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -2,147 +2,80 @@
 import { ref, computed } from 'vue';
 import { cleanDescription } from '../../utils/CleanDescription';
 import { generateNGrams } from '../../utils/generateNGrams';
+import { MaTextarea, MaSelect2 as MaSelect, MaBadge, MaButton, MaCheckbox2 as MaCheckbox, MaCard, MaEmpty } from '@mobileaction/action-kit';
 
 const sourceDescriptionText = ref('');
+const selectedNGrams = ref([1, 2, 3]);
+const shouldRemoveStopWords = ref(true);
+const maxNumberForNgram = 10;
```
sevvalkaplannn: constants should follow uppercase naming convention.
```suggestion
const MAX_NGRAM_SIZE = 10;
```

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -2,147 +2,80 @@
...
+
+const generatedKeywordNGrams = computed(() => {
     const text = sourceDescriptionText.value;
     if (!text.trim()) return [];
 
-    const cleanedText = cleanDescription(text);
+    const cleanedText = cleanDescription(text, shouldRemoveStopWords.value);
     
-    return generateNGrams(cleanedText);
+    const sortedNGrams = [...selectedNGrams.value].sort((a, b) => a - b);
```
sevvalkaplannn: this runs every time the generatedKeywordNGrams computed property evaluates, since sorting is a concern of the selected values themselves we can consider creating a dedicated computed for this

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -2,147 +2,80 @@
...
+const maxNumberForNgram = 10;
 
-const generatedKeywordNGrams = computed(() => 
-{
+const nGramOptions = Array.from({ length: maxNumberForNgram }, (_, i) => ({
+    value: i + 1,
+    label: `${i + 1}-Gram`
+}));
+
+const generatedKeywordNGrams = computed(() => {
```
sevvalkaplannn: this computed currently contains multiple responsibilities, we can consider extracting the jobs into multiple computeds

src/pages/keyword-generator/KeywordGenerator.vue
sevvalkaplannn: now we are skipping of unavailable n-grams, prefer returning empty groups this keeps UI predictable, and you can use MaEmpty with that case

sevvalkaplannn (changes requested):

sevvalkaplannn (approved):
thanks for your work, you can continue with ONB-203 🙌🏻

PR #108, opened 2026-06-26
feat(ONB-203): Keywords are found through a button.

okayhamamci:
Navbar is added, custom CSS are replaced with tailwindcss.


sevvalkaplannn (changes requested):
1- this commit is not a feat commit, it is chore, can you check the [document](https://mobileaction.atlassian.net/wiki/spaces/Engineerin/pages/1456243537/Conventional+Commit+Messages) for your next commits.

2- Can you add more ui improvements into your navbar.

3- You can add more ui improvements, for example icons for buttons, and the buttons should be seen at first even if the input is not written, but they can be disabled until user enters the input, the notifications can be more descriptive, you can add notification for clear text button, even if you want you can add MaPopover2 into your buttons for validation.

General aim of this task is improving the ui and be more creative in case of stylings and actionkit usages.

sevvalkaplannn (commented):
can you improve your navbar, you can check some examples

src/pages/keyword-generator/KeywordGenerator.vue
```suggestion
const sourceDescriptionText = ref('');
const selectedNGrams = ref([1, 2, 3]);
const shouldRemoveStopWords = ref(true);
```

refs can be camelCase

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -23,17 +23,64 @@ const cleanedSourceText = computed(() => {
...
+    if (!cleanedSourceText.value) {
+        MaNotification.error({
+            title: 'No Keywords After Cleaning',
+            message: SHOULD_REMOVE_STOP_WORDS.value
+                ? 'Your text only contains stop words or unsupported characters. Add more descriptive words or turn off stop word removal.'
+                : 'Your text only contains unsupported characters. Add words or phrases before generating keywords.'
+        });
+        generatedKeywordNGrams.value = [];
+        return;
+    }
```
sevvalkaplannn, line 34: generateKeywordsOnDemand function has too many jobs to do, we can separate the functionalities

function validateInput() {}

function generateKeywords() {}

function showSuccessNotification() {}

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -23,17 +23,64 @@ const cleanedSourceText = computed(() => {
...
+const totalGeneratedKeywordCount = computed(() => generatedKeywordNGrams.value.reduce((total, nGramCategory) => total + nGramCategory.keywords.length, 0));
+const selectedNGramLabels = computed(() => sortedSelectedNGrams.value.map(nGram => `${nGram}-gram`).join(', '));
+
+const generateKeywordsOnDemand = () => {
+    if (!hasInput.value) {
+        MaNotification.error({
+            title: 'Text Required',
+            message: 'Paste an app description or keyword list before generating keyword ideas.'
+        });
+        generatedKeywordNGrams.value = [];
```
sevvalkaplannn: too many repetition of  generatedKeywordNGrams.value = []; instead of that we can create and use;

```
const clearResults = () => {
    generatedKeywordNGrams.value = [];
};
``

and call clearResults(); to avoid repetition

File: src/pages/keyword-generator/KeywordGenerator.vue [hunk 3]
```diff
@@ -23,17 +23,64 @@ const cleanedSourceText = computed(() => {
     return cleanDescription(text, SHOULD_REMOVE_STOP_WORDS.value);
 });
```
sevvalkaplannn: we can simplify it
```suggestion
const cleanedSourceText = computed(() =>
    SOURCE_DESCRIPTION_TEXT.value.trim()
        ? cleanDescription(
              SOURCE_DESCRIPTION_TEXT.value,
              SHOULD_REMOVE_STOP_WORDS.value
          )
        : ''
);
```

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -62,12 +109,32 @@ const hasInput = computed(() => SOURCE_DESCRIPTION_TEXT.value.trim().length > 0)
...
+                        Clear Text
+                    </MaButton>
+                    <MaButton class="w-48" type="primary" :disabled="!hasInput" @click="generateKeywordsOnDemand">
+                        <template #icon>
+                            <svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" stroke-width="1.5" stroke="currentColor" class="w-4 h-4 mr-1">
+                                <path stroke-linecap="round" stroke-linejoin="round" d="M3.75 13.5l10.5-11.25L12 10.5h8.25L9.75 21.75 12 13.5H3.75z" />
+                            </svg>
+                        </template>
+                        Generate Keywords
+                    </MaButton>
```
sevvalkaplannn: don't need the icon slot, MaButton already has icon prop and many icons in it please check the [doc](https://mobileaction.github.io/action-kit/?path=/docs/core-mabutton-default-filled--docs)

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -88,5 +155,10 @@ const hasInput = computed(() => SOURCE_DESCRIPTION_TEXT.value.trim().length > 0)
                 </MaEmpty>
             </MaCard>
         </div>
+        <MaEmpty v-else description="Generated keywords will appear here after you enter text and run the generator.">
+            <template #icon>
+                <span class="hidden"></span>
+            </template>
+        </MaEmpty>
```
sevvalkaplannn: MaEmpty has its on icons please check [actionkit](https://mobileaction.github.io/action-kit/?path=/docs/advanced-maempty--docs)

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -62,12 +109,32 @@ const hasInput = computed(() => SOURCE_DESCRIPTION_TEXT.value.trim().length > 0)
...
+                </p>
+                <div class="flex shrink-0 flex-row gap-3">
+                    <MaButton class="w-36" :disabled="!hasInput" @click="resetKeywordInput">
+                        <template #icon>
+                            <svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" stroke-width="1.5" stroke="currentColor" class="w-4 h-4 mr-1">
+                                <path stroke-linecap="round" stroke-linejoin="round" d="M6 7.5h12M9.75 7.5v-.75A1.5 1.5 0 0 1 11.25 5.25h1.5a1.5 1.5 0 0 1 1.5 1.5v.75m-6.75 3v7.5m4.5-7.5v7.5m4.5-7.5v7.5M7.5 7.5l.75 12A1.5 1.5 0 0 0 9.75 21h4.5a1.5 1.5 0 0 0 1.5-1.5l.75-12" />
+                            </svg>
+                        </template>
+                        Clear Text
+                    </MaButton>
```
sevvalkaplannn: please check and fix the all MaButtons in your project, you don't need to put icon manually, we have prop

sevvalkaplannn (changes requested):
1- if textarea is not changed we can disable the generate button since the array is stale.
2- when I click the button a scrollbar appears on x axis, it is caused because of the css I guess,  can you find and fix the css that cause that issue

https://github.com/user-attachments/assets/3535b2ef-0ede-44b2-93a6-aa6166790934

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,39 +1,88 @@
...
+
+const hasInput = computed(() => sourceDescriptionText.value.trim().length > 0);
+const hasSelectedNGrams = computed(() => sortedSelectedNGrams.value.length > 0);
+const hasGeneratedKeywords = computed(() => generatedKeywordNGrams.value.length > 0);
+const totalGeneratedKeywordCount = computed(() => generatedKeywordNGrams.value.reduce((total, nGramCategory) => total + nGramCategory.keywords.length, 0));
+const selectedNGramLabels = computed(() => sortedSelectedNGrams.value.map(nGram => `${nGram}-gram`).join(', '));
+
+const lastGeneratedState = ref(null);
+
+const isStale = computed(() => {
```
```suggestion
const hasUnchangedInputs = computed(() => {
```
naming can be hasUnchangedInputs

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,39 +1,88 @@
...
+const shouldRemoveStopWords = ref(true);
 const MAX_NGRAM_SIZE = 10;
 
 const nGramOptions = Array.from({ length: MAX_NGRAM_SIZE }, (_, i) => ({
     value: i + 1,
     label: `${i + 1}-Gram`
 }));
 
-const sortedSelectedNGrams = computed(() => [...SELECTED_NGRAMS.value].sort((a, b) => a - b));
+const sortedSelectedNGrams = computed(() => [...selectedNGrams.value].sort((a, b) => a - b));
```
sevvalkaplannn: it would be cleaner to normalize only where ordering actually matters instead of making every consumer use a sorted version, because now it is used in every time when the selection changes

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,39 +1,88 @@
...
     
-    return generateNGrams(cleanedSourceText.value, sortedSelectedNGrams.value);
-});
+    generatedKeywordNGrams.value = generateKeywords(cleanedSourceText.value, sortedSelectedNGrams.value);
+    
+    lastGeneratedState.value = {
+        text: sourceDescriptionText.value,
+        nGrams: sortedSelectedNGrams.value.join(','),
+        removeStopWords: shouldRemoveStopWords.value
+    };
```
sevvalkaplannn: string serialization as comparison logic isn't ideal, store the actual array and compare arrays

src/utils/keywordGeneratorActions.js
```diff
@@ -0,0 +1,43 @@
...
+        });
+        return false;
+    }
+
+    return true;
+}
+
+export function generateKeywords(cleanedSourceText, sortedSelectedNGrams) {
+    return generateNGrams(cleanedSourceText, sortedSelectedNGrams);
+}
```
sevvalkaplannn: generateKeywords() utility adds almost no value, this wrapper doesn't add abstraction, so we do not need it, it just forwards the parameters now, you can call generateNGrams directly.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,39 +1,88 @@
...
-    
-    return cleanDescription(text, SHOULD_REMOVE_STOP_WORDS.value);
+const cleanedSourceText = computed(() =>
+    sourceDescriptionText.value.trim()
+        ? cleanDescription(
+            sourceDescriptionText.value,
+            shouldRemoveStopWords.value
+        )
+        : ''
+);
```
sevvalkaplannn: cleaning happens on every call,  however we need it only needed when user clicks the generate button, it would be better performed inside the generation flow instead of continuously

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,39 +1,88 @@
...
+
+const generatedKeywordNGrams = ref([]);
+
+const hasInput = computed(() => sourceDescriptionText.value.trim().length > 0);
+const hasSelectedNGrams = computed(() => sortedSelectedNGrams.value.length > 0);
+const hasGeneratedKeywords = computed(() => generatedKeywordNGrams.value.length > 0);
+const totalGeneratedKeywordCount = computed(() => generatedKeywordNGrams.value.reduce((total, nGramCategory) => total + nGramCategory.keywords.length, 0));
+const selectedNGramLabels = computed(() => sortedSelectedNGrams.value.map(nGram => `${nGram}-gram`).join(', '));
+
+const lastGeneratedState = ref(null);
```
sevvalkaplannn: lastGeneratedState exists just to disable the button, this introduces extra synchronization responsibility, you don't need to duplicate the state, you can use the first keywords array directly

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,39 +1,88 @@
...
+const isStale = computed(() => {
+    if (!lastGeneratedState.value) return false;
+    return lastGeneratedState.value.text === sourceDescriptionText.value &&
+           lastGeneratedState.value.nGrams === sortedSelectedNGrams.value.join(',') &&
+           lastGeneratedState.value.removeStopWords === shouldRemoveStopWords.value;
 });
 
-const generatedKeywordNGrams = computed(() => {
-    if (!cleanedSourceText.value) return [];
+const clearResults = () => {
```
sevvalkaplannn: don't need this function, you can use generatedKeywordNGrams.value = []; in places you need, its not a necessary abstraction

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -43,31 +92,41 @@ const hasInput = computed(() => SOURCE_DESCRIPTION_TEXT.value.trim().length > 0)
             <p class="text-gray-500 text-base m-0">Generate 1-10 gram keywords from your text without duplicates.</p>
```
sevvalkaplannn: you can use MAX_NGRAM_SIZE in this template since if the max changes later, someone has to remember to update the text

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,39 +1,88 @@
...
+    
+    showSuccessNotification(totalGeneratedKeywordCount.value, selectedNGramLabels.value);
+};
 
 const resetKeywordInput = () => {
-    SOURCE_DESCRIPTION_TEXT.value = '';
-};
+    const hadGeneratedKeywords = hasGeneratedKeywords.value;
+    sourceDescriptionText.value = '';
+    clearResults();
```
```suggestion
generatedKeywordNGrams.value = [];
```
don't need to use clearResults, we can delete that function since it is not necessarily complicated

src/pages/keyword-generator/KeywordGenerator.vue
sevvalkaplannn: [MaCard](https://mobileaction.github.io/action-kit/?path=/docs/core-macard--docs) has title prop and headerActions slot, can you check you can use them

src/utils/keywordGeneratorActions.js
```diff
@@ -0,0 +1,43 @@
...
+            title: 'No Keywords After Cleaning',
+            message: shouldRemoveStopWords
+                ? 'Your text only contains stop words or unsupported characters. Add more descriptive words or turn off stop word removal.'
+                : 'Your text only contains unsupported characters. Add words or phrases before generating keywords.'
+        });
+        return false;
+    }
+
+    return true;
+}
```
sevvalkaplannn: the validation flow is more complex than necessary, because it requires passing several values (hasInput, hasSelectedNGrams, cleanedSourceText, shouldRemoveStopWords) instead of validating the original inputs directly, a simpler approach would be to validate the raw input values sequentially, making the code easier to read and reducing the number of intermediate computed properties, now it looks to complicated

sevvalkaplannn (changes requested):
on generateNGrams function in `@src/utils/generateNGrams.js` file you used `cleanedText.split(' ')`, it is a simplistic tokenizer but it doesn't cover the tabs, newlines and unicode whitespace so it would be better to use it like `cleanedText.trim().split(/\s+/)`

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,78 +1,122 @@
...
+                        Clear Text
+                    </MaButton>
+                    <MaButton class="w-48" color="dark" variant="stroke" type="primary" :disabled="!hasInput" @click="generateKeywordsOnDemand">
+                        <template #icon>
+                            <svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" stroke-width="1.5" stroke="currentColor" class="w-4 h-4 mr-1">
+                                <path stroke-linecap="round" stroke-linejoin="round" d="M3.75 13.5l10.5-11.25L12 10.5h8.25L9.75 21.75 12 13.5H3.75z" />
+                            </svg>
+                        </template>
+                        Generate Keywords
+                    </MaButton>
```
sevvalkaplannn: MaButton has its own icons you don't need to add custom svg to your button, can you check your all actionkit usages.

sevvalkaplannn (changes requested):

src/components/TheNav.vue
```diff
@@ -1,16 +1,74 @@
...
+            <div class="flex items-center h-6 w-24">
+                <MaImage2
+                    src="/ma.png"
+                    alt="MobileAction logo"
+                    width="100%"
+                    height="100%"
+                    fit="contain"
+                />
+            </div>
+        </div>
```
sevvalkaplannn: there are unnecessary div's we don't need both of them

src/components/TheNav.vue
```diff
@@ -1,16 +1,74 @@
...
+                    Home
+                </router-link>
+                <router-link to="/keyword-generator" class="whitespace-nowrap rounded-md px-3 py-1.5 text-sm text-black hover:bg-gray-50 hover:text-blue-700 font-medium transition-colors">
+                    Keyword Generator
+                </router-link>
+                <router-link to="/keyword-density" class="whitespace-nowrap rounded-md px-3 py-1.5 text-sm text-black hover:bg-gray-50 hover:text-blue-700 font-medium transition-colors">
+                    Keyword Density
+                </router-link>
+            </nav>
+        </div>
```
sevvalkaplannn: you don't need to wrap the nav with div

src/components/TheNav.vue
```diff
@@ -1,16 +1,74 @@
...
+        <div class="flex items-center justify-center sm:justify-end">
+            <MaButton size="small" @click="isNavigationDrawerOpen = true">
+                Menu
+            </MaButton>
+        </div>
```
sevvalkaplannn: you don't need this div too, the alignment in this navbar should handled by the parent wrapper, now all of them are individual with this implementation

src/components/TheNav.vue
```diff
@@ -1,16 +1,74 @@
...
+        <MaDrawer
```
sevvalkaplannn, line 35: both navbar menu and drawer menu are unnecessary together, the drawer can be seen on small screens, but in larger screens we don't need both of them together

src/components/TheNav.vue
```diff
@@ -1,16 +1,74 @@
...
+    <div class="m-4 grid grid-cols-1 items-center gap-4 rounded-lg border border-gray-100 bg-white px-4 py-1 shadow-sm sm:grid-cols-[150px_1fr_150px] sm:gap-0">
+        <div class="flex justify-center sm:justify-start">
+            <div class="flex items-center h-6 w-24">
+                <MaImage2
+                    src="/ma.png"
+                    alt="MobileAction logo"
+                    width="100%"
+                    height="100%"
+                    fit="contain"
+                />
```
sevvalkaplannn: MaImage2 component is not suitable for navbar icon, you can add the svg into your project as mobileaction.svg and then import it here.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,78 +1,117 @@
...
+const hasGeneratedKeywords = computed(() => generatedKeywordNGrams.value.length > 0);
+const totalGeneratedKeywordCount = computed(() => generatedKeywordNGrams.value.reduce((total, nGramCategory) => total + nGramCategory.keywords.length, 0));
 
-const cleanedSourceText = computed(() => {
-    const text = SOURCE_DESCRIPTION_TEXT.value;
-    if (!text.trim()) return '';
-    
-    return cleanDescription(text, SHOULD_REMOVE_STOP_WORDS.value);
-});
+const generateKeywordsOnDemand = () => {
```
sevvalkaplannn, line 20: still generateKeywordsOnDemand is doing too much, thats a lot of responsibilities, it would be cleaner to extract the generation flow into a pure utility, then the component only updates refs and displays notifications

the current implementation mixes ui logic with algorithm logic, this Vue component should be responsible for interacting with the ui (notifications, updating refs), while utility functions should focus on pure business logic such as validation, text processing, and keyword generation. keeping these responsibilities separate makes the code easier to test, reuse, and maintain.

src/utils/keywordGeneratorActions.js
```diff
@@ -0,0 +1,37 @@
+import { MaNotification } from '@mobileaction/action-kit';
+
+export function validateInput(sourceDescriptionText, selectedNGrams, cleanedSourceText, shouldRemoveStopWords) {
```
sevvalkaplannn, line 4: validateInput() currently performs validation and displays notifications, instead, the utility should only determine whether the input is valid and return the validation result. The component should decide how that result is presented to the user, so we should call these MaNotifications on vue component,

sevvalkaplannn (changes requested):

src/components/TheNav.vue
```diff
@@ -1,16 +1,79 @@
...
-        </router-link>
+    <div class="m-4 grid grid-cols-1 items-center gap-4 rounded-lg border border-gray-100 bg-white px-4 py-1 shadow-sm sm:grid-cols-[150px_1fr_150px] sm:gap-0">
+        <div class="flex h-6 w-24 items-center justify-self-center sm:justify-self-start">
+            <img
+                src="/mobileaction.svg"
+                alt="MobileAction logo"
+                class="w-full h-full object-contain"
+            />
+        </div>
+        <nav class="desktop-nav flex w-full flex-row flex-nowrap justify-center gap-2 sm:gap-4 items-center overflow-x-auto">
```
sevvalkaplannn:
```suggestion
        <nav class="flex w-full flex-row flex-nowrap justify-center gap-2 sm:gap-4 items-center overflow-x-auto hidden sm:block">
```

with this tailwind classes you don't need below css.

src/components/TheNav.vue
```diff
@@ -1,16 +1,79 @@
...
+                Home
+            </router-link>
+            <router-link to="/keyword-generator" class="whitespace-nowrap rounded-md px-3 py-1.5 text-sm text-black hover:bg-gray-50 hover:text-blue-700 font-medium transition-colors">
+                Keyword Generator
+            </router-link>
+            <router-link to="/keyword-density" class="whitespace-nowrap rounded-md px-3 py-1.5 text-sm text-black hover:bg-gray-50 hover:text-blue-700 font-medium transition-colors">
+                Keyword Density
+            </router-link>
+        </nav>
+        <MaButton class="mobile-menu-btn justify-self-center" size="small" @click="isNavigationDrawerOpen = true">
```
sevvalkaplannn:
```suggestion
        <MaButton class="justify-self-center block sm:hidden" size="small" @click="isNavigationDrawerOpen = true">
```
with this tailwind classes you don't need below css

src/components/TheNav.vue
```diff
@@ -1,16 +1,79 @@
...
+    .desktop-nav {
+        display: none !important;
+    }
+}
+@media (min-width: 640px) {
+    .mobile-menu-btn {
+        display: none !important;
+    }
+}
 </style>
```
sevvalkaplannn: we don't need this part, I suggested the tailwind codes for this part
```suggestion

```

sevvalkaplannn (approved):

PR #109, opened 2026-06-29
ONB-204: add keyword density page

cmlcz:
Adds a Keyword Density page that shows each keyword's count and density (% of total words) in a table.


sevvalkaplannn (changes requested):
in this task we need to make our project responsive, can you improve the responsive ui too for all pages, you can make your navbar responsive too (maybe it can be hamburger menu)

sevvalkaplannn (changes requested):
ı have some ui/ux comments

1- the navbar is transparent right now and its not that ui/ux friendly

https://github.com/user-attachments/assets/54c900b3-2f15-490c-bc85-b58c74f19300

2- Small screen ui is very good, but for making small screen ui better you changed the larger screen ui and it was better before. And you changed the navbar to only hamburger menu. This task's purpose is implementing a project for all resolutions, but their ui can be differ, for example for small screens use hamburger menu but for larger screens you can keep the old navbar.

3- The components are full width now, but your old ui was better, for only smaller screens you can keep this ui but for larger screens full width components are not ui friendly. The aim of responsive design is being able to implement and create different ui's with css only.

https://github.com/user-attachments/assets/bceb6a49-6f9c-4d37-b4b1-1706e74d56d8

src/components/TheHeader.vue
```diff
@@ -1,9 +1,19 @@
...
 import TheNav from './TheNav.vue';
+
+const mediaQuery = window.matchMedia('(min-width: 768px)');
+const isDesktop = ref(mediaQuery.matches);
+const syncIsDesktop = () => {
+    isDesktop.value = mediaQuery.matches;
+};
+
+onMounted(() => mediaQuery.addEventListener('change', syncIsDesktop));
+onBeforeUnmount(() => mediaQuery.removeEventListener('change', syncIsDesktop));
```
sevvalkaplannn: thanks for your work getting the responsive behavior working, but this doesn't need javascript to detect the screen size. since you're already using tailwind, you can let css handle the responsive layout with its breakpoint utilities like md, lg...

for example, you can render both the desktop navigation and the mobile menu, then use classes like hidden md:flex for the desktop version and flex md:hidden for the mobile version. that removes the need for window.matchMedia, the isDesktop prop, event listeners, and the watcher, making the component simpler and easier to maintain.

src/components/TheNav.vue
```diff
@@ -1,48 +1,70 @@
...
+const activeLinkClass = 'bg-gray-900 text-white shadow-md';
+
 const isOpen = ref(false);
+
+watch(
+    () => props.isDesktop,
+    (isDesktop) => {
+        if (isDesktop) isOpen.value = false;
+    },
+);
```
sevvalkaplannn: javascript should generally only be used when the viewport size affects application logic, in this case, it's only affecting presentation, so tailwinds responsive utilities are the more suitable and enough for this project.

src/pages/keyword-density/KeywordDensity.vue
sevvalkaplannn: I don't think calculateDensity() should be called at the top level of the component, since this is only for the initial state, initialize rows and lastCalculated directly when declaring them. this calculateDensity function should be handled for the button click handler.

sevvalkaplannn (changes requested):

src/components/TheHeader.vue
cmlcz: Yeah you're right, that's the cleaner way to do it. I actually tried the hidden md:flex setup first but the md breakpoints weren't compiling in my local tailwind and vite setup. So the desktop nav never showed up on my end and I went with hamburger navbar for this time. If I figure out what's the problem or what's going on with my environment, I'll switch it

sevvalkaplannn: If you have any questions we can make a huddle 🙌🏻

src/utils/KeywordDensity.js
```diff
@@ -0,0 +1,31 @@
...
+    const countByKeyword = {};
+    for (const word of allWords) {
+        if (stopWords.includes(word)) continue;
+        countByKeyword[word] = (countByKeyword[word] || 0) + 1;
+    }
+
+    const keywordsByCount = {};
+    for (const [keyword, count] of Object.entries(countByKeyword)) {
+        (keywordsByCount[count] ??= []).push(keyword);
+    }
```
sevvalkaplannn: keywordsByCount object makes the implementation a bit harder to follow, since the final output is just grouped by count, you can use map instead of creating two separate objects

sevvalkaplannn (commented):
good work, I just left a little suggestion. By the way your menu looks very good now, you don't need to change it, just you can rename the file name since its not navbar anymore its menu.

src/components/TheHeader.vue
cmlcz: I'll take another look at it, after that we can make a huddle

sevvalkaplannn (approved):
good work 🙌🏻

PR #111, opened 2026-07-01
ONB-205: add light/dark theme and pinia stores

cmlcz:
Adds a light/dark theme toggle and introduces Pinia for shared state. The input text is now shared between the Keyword Generator and Keyword Density pages through a store.

### DIFF
--- a/src/components/TheHeader.vue
+++ b/src/components/TheHeader.vue
@@ -1,6 +1,8 @@
 <script setup>
 import TheMenu from './TheMenu.vue';
 import { useThemeStore } from '@/stores/theme';
+import sunIcon from '@/assets/sun.svg';
+import moonIcon from '@/assets/moon.svg';
 
 const theme = useThemeStore();
 </script>
@@ -14,13 +16,7 @@ const theme = useThemeStore();
             class="flex h-11 w-11 items-center justify-center rounded-full border border-gray-200/80 bg-white/80 text-gray-700 shadow-lg backdrop-blur-md transition hover:bg-gray-100 active:scale-95"
             @click="theme.toggle"
         >
-            <svg v-if="theme.isDark" class="w-5 h-5" fill="none" stroke="currentColor" stroke-width="2" viewBox="0 0 24 24">
-                <circle cx="12" cy="12" r="4"/>
-                <path stroke-linecap="round" d="M12 2v2M12 20v2M4.93 4.93l1.41 1.41M17.66 17.66l1.41 1.41M2 12h2M20 12h2M6.34 17.66l-1.41 1.41M19.07 4.93l-1.41 1.41"/>
-            </svg>
-            <svg v-else class="w-5 h-5" fill="none" stroke="currentColor" stroke-width="2" viewBox="0 0 24 24">
-                <path stroke-linecap="round" stroke-linejoin="round" d="M21 12.79A9 9 0 1 1 11.21 3 7 7 0 0 0 21 12.79z"/>
-            </svg>
+            <img :src="theme.isDark ? sunIcon : moonIcon" class="w-5 h-5" alt="">
         </button>
     </header>
 </template>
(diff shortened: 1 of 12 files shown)


sevvalkaplannn (commented):
1- I think 1, 2 and 3-gram options can be selected as default. 

https://github.com/user-attachments/assets/f6c90951-575f-4acb-8dc4-87515e3e6c0e

2- When user selects 2,7 and 8-gram, we should show all of the selected n-gram in the keyword generator page, for example in the below image we can show 2 7 and 8 on the results part but we can write empty text for 7 and 8, in this current version we are seeing 3 options as selected but in result we only see one.

Image

sevvalkaplannn (changes requested):
1- there is a bug on keyword-generator page, we are not able to click the generate button on first load
https://github.com/user-attachments/assets/1fbb7161-e426-49fc-ae70-bef4115dbf83

sevvalkaplannn (commented):
you can add "No n-gram result" text or another text to expanded version of empty n-gram's for example when we expanded the 7 and 10gram we should see empty case text

Image

src/components/TheHeader.vue
```diff
@@ -1,9 +1,26 @@
...
+            class="flex h-11 w-11 items-center justify-center rounded-full border border-gray-200/80 bg-white/80 text-gray-700 shadow-lg backdrop-blur-md transition hover:bg-gray-100 active:scale-95"
+            @click="theme.toggle"
+        >
+            <svg v-if="theme.isDark" class="w-5 h-5" fill="none" stroke="currentColor" stroke-width="2" viewBox="0 0 24 24">
+                <circle cx="12" cy="12" r="4"/>
+                <path stroke-linecap="round" d="M12 2v2M12 20v2M4.93 4.93l1.41 1.41M17.66 17.66l1.41 1.41M2 12h2M20 12h2M6.34 17.66l-1.41 1.41M19.07 4.93l-1.41 1.41"/>
+            </svg>
+            <svg v-else class="w-5 h-5" fill="none" stroke="currentColor" stroke-width="2" viewBox="0 0 24 24">
+                <path stroke-linecap="round" stroke-linejoin="round" d="M21 12.79A9 9 0 1 1 11.21 3 7 7 0 0 0 21 12.79z"/>
+            </svg>
```
sevvalkaplannn: can you add these svg's into your project's assets folder, after doing it you can use here by importing those files

src/stores/theme.js
```diff
@@ -0,0 +1,15 @@
...
+
+export const useThemeStore = defineStore('theme', () => {
+    const isDark = ref(localStorage.getItem('theme') === 'dark');
+
+    watch(isDark, (dark) => {
+        document.documentElement.classList.toggle('dark', dark);
+        localStorage.setItem('theme', dark ? 'dark' : 'light');
+    }, { immediate: true });
+
+    const toggle = () => (isDark.value = !isDark.value);
```
sevvalkaplannn:
```suggestion
    const toggle = () => (isDark.value = !isDark.value);

    watch(isDark, (dark) => {
        document.documentElement.classList.toggle('dark', dark);
        localStorage.setItem('theme', dark ? 'dark' : 'light');
    }, { immediate: true });
```
in the composition API, the execution order of the code depends on where it is defined. To improve readability and ensure smooth operation within the `<script setup>` block, the logical order from top to bottom is generally `Import -> Props/Emits -> State -> Computed -> Methods -> Watch -> Lifecycle Hooks`

sevvalkaplannn (changes requested):
1- Can you add copy to clipboard button into your keyword-density table, it should give us csv formatted table data.

sevvalkaplannn (approved):
good work 🚀

sevvalkaplannn (commented):
fyi your last commit is not a feat commit, it is chore too. Feat is more like a commit tag for bigger improvements

PR #112, opened 2026-07-02
feat(ONB-205): Pinia store is implemented.

okayhamamci:
Input of density and generator page matches from now on with these minor details. If user does not any input to the keyword generator, constant text appears on keyword density. If user passes to keyword generator again without writing anything to the keyword density, input area of keyword generator appears empty instead of constant text on the keyword density.


sevvalkaplannn (changes requested):
can you make dark/light theme into your project by using pinia store too, by doing it we can improve the ui of the project and its a good practice to use pinia too 🙌🏻

src/pages/keyword-density/KeywordDensity.vue
```diff
@@ -1,12 +1,14 @@
...
+import { storeToRefs } from 'pinia';
+import { useKeywordStore } from '../../stores/keywordStore';
 import { MaTextarea, MaButton, MaCheckbox2 as MaCheckbox, MaNotification } from '@mobileaction/action-kit';
 import { processKeywordDensity } from '../../utils/calculateDensity';
 import { AgGridVue } from 'ag-grid-vue3';
 
-const STATIC_PARENT_TEXT = 'Our Keyword Counter tool lets you count how many times keywords are repeated in any text, and also calculates the density of these keywords. The keyword density is the percentage of times a keyword appears in a text compared to the total number of words in that text. Simply write or paste your text here and hit "count".';
-
-const inputText = ref(STATIC_PARENT_TEXT);
+const store = useKeywordStore();
```
sevvalkaplannn:
```suggestion
const keywordStore = useKeywordStore();
```

src/pages/keyword-density/KeywordDensity.vue
```diff
@@ -1,12 +1,14 @@
 <script setup>
 import { ref, computed } from 'vue';
+import { storeToRefs } from 'pinia';
+import { useKeywordStore } from '../../stores/keywordStore';
```
sevvalkaplannn:
```suggestion
import { useKeywordStore } from '@/stores/keywordStore';
```
we can use the imports with .../ parent directory paths but this full path usage is more readable, can you change your imports of all of the files in this project to this module path alias usages.

src/pages/keyword-density/KeywordDensity.vue
sevvalkaplannn: naming is everything for us, in some components we need to use more than one store, so for that cases we need to get used to writing everything readable and understandable

src/pages/keyword-density/KeywordDensity.vue
```diff
@@ -1,12 +1,14 @@
...
 import { MaTextarea, MaButton, MaCheckbox2 as MaCheckbox, MaNotification } from '@mobileaction/action-kit';
 import { processKeywordDensity } from '../../utils/calculateDensity';
 import { AgGridVue } from 'ag-grid-vue3';
 
-const STATIC_PARENT_TEXT = 'Our Keyword Counter tool lets you count how many times keywords are repeated in any text, and also calculates the density of these keywords. The keyword density is the percentage of times a keyword appears in a text compared to the total number of words in that text. Simply write or paste your text here and hit "count".';
-
-const inputText = ref(STATIC_PARENT_TEXT);
+const store = useKeywordStore();
+store.initDensityText();
+const { sharedInputText: inputText } = storeToRefs(store);
```
sevvalkaplannn: storeToRefs is not needed here because the store only exposes a single ref (sharedInputText), in a pinia setup store, refs are already reactive and are automatically unwrapped when accessed through the store, so you can simply use 

const inputText = keywordStore.sharedInputText;

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,8 +1,16 @@
 <script setup>
 import { ref, computed } from 'vue';
+import { storeToRefs } from 'pinia';
```
```suggestion
```
its not used in this file

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,8 +1,16 @@
...
+import { useKeywordStore } from '../../stores/keywordStore';
 import { processKeywords } from '../../utils/keywordGeneratorActions';
 import { MaTextarea, MaSelect2 as MaSelect, MaBadge, MaButton, MaCheckbox2 as MaCheckbox, MaCard, MaEmpty, MaNotification } from '@mobileaction/action-kit';
-const sourceDescriptionText = ref('');
+
+const store = useKeywordStore();
+const sourceDescriptionText = computed({
+    get: () => store.sharedInputText === store.STATIC_PARENT_TEXT ? '' : store.sharedInputText,
+    set: (val) => { store.sharedInputText = val; }
+});
```
sevvalkaplannn: the computed getter/setter isn't needed here. sharedInputText is already a writable ref in the pinia store, so you can bind it directly to the textarea instead of wrapping it in a computed property.


```suggestion
const sourceDescriptionText = store.sharedInputText;
```

src/stores/keywordStore.js
```diff
@@ -0,0 +1,20 @@
...
+    function initDensityText() {
+        if (!sharedInputText.value) {
+            sharedInputText.value = STATIC_PARENT_TEXT;
+        }
+    }
+
+    return {
+        sharedInputText,
+        STATIC_PARENT_TEXT,
+        initDensityText
```
sevvalkaplannn: you don't need to return what you won't gonna use in the components, as far as I see you only use sharedInputText, so you can delete the other returns
```suggestion
        sharedInputText,
```

sevvalkaplannn (changes requested):
1- Below empty case can be MaEmpty from actionkit.

Image

sevvalkaplannn (commented):
this commit is not a feat commit, its a refactor fyi

src/pages/keyword-density/KeywordDensity.vue
```diff
@@ -73,35 +96,34 @@ const copyToClipboard = async () => {
...
                 </div>
             </div>
 
             <div class="w-full lg:w-1/2 flex flex-col gap-4">
-                <div v-if="keywordStats.length === 0" class="border border-gray-200 rounded-lg p-8 text-center text-gray-500 text-sm bg-white shadow-sm">
-                    No keywords to display. Enter text and calculate density.
-                </div>
-                <div v-else class="ag-theme-alpine w-full shadow-sm rounded-lg overflow-hidden border border-gray-200">
+                <MaEmpty v-if="keywordStats.length === 0" description="No keywords to display. Enter text and calculate density." class="dark:invert dark:hue-rotate-180"></MaEmpty>
+                <div v-else class="ag-theme-alpine w-full shadow-sm rounded-lg overflow-hidden border border-gray-200 dark:invert dark:hue-rotate-180">
```
sevvalkaplannn:
```suggestion
                <div v-else class="w-full shadow-sm rounded-lg overflow-hidden border border-gray-200 dark:invert dark:hue-rotate-180">
```

sevvalkaplannn (changes requested):
Instead of adding dark css to each components, you can change your src/main.css file to behave for dark theme, in this version like in below video there are some problems with empty cases and above the header. 
 
https://github.com/user-attachments/assets/f0b7291f-7b79-4884-82c4-435c956ed008

sevvalkaplannn (commented):
Instead of dark/light button you can use [MaSwitch](https://mobileaction.github.io/action-kit/?path=/docs/core-maswitch--docs) from actionkit

src/pages/keyword-density/KeywordDensity.vue
```diff
@@ -1,17 +1,35 @@
...
 const shouldRemoveStopWords = ref(false);
-const totalCharacters = computed(() => inputText.value.length);
-const hasInput = computed(() => inputText.value.trim().length > 0);
+const totalCharacters = computed(() => keywordStore.sharedInputText.length);
+const hasInput = computed(() => keywordStore.sharedInputText.trim().length > 0);
+
+const lastCalculatedState = ref({
+    text: null,
+    removeStopWords: null
+});
```
sevvalkaplannn:
```suggestion
});

```

src/style.css
```diff
@@ -1 +1,72 @@
...
+
+.ma-card h1,
+.ma-card h2,
+.ma-card h3,
+.ma-card h4,
+.ma-card h5,
+.ma-card h6,
+.ma-card [class*="title"] {
+    @apply !text-black;
+}
```
sevvalkaplannn:
```suggestion
html.dark {
    background: #fff;
    filter: invert(1) hue-rotate(180deg);
}
```
you can simply apply it like that

src/components/TheNav.vue
```diff
@@ -1,16 +1,33 @@
...
+];
+
+const currentTheme = computed({
+    get: () => themeStore.isDarkMode ? 'dark' : 'light',
+    set: (val) => {
+        if ((val === 'dark') !== themeStore.isDarkMode) {
+            themeStore.toggleTheme();
+        }
+    }
+});
```
sevvalkaplannn: this computed get/set wrapper isn't needed, whole point of a pinia store is to be the single source of truth you read/write directly wrapping it in a local computed just to translate 'dark'/'light' to boolean adds a layer that has to stay in sync with the store for no benefit. Instead of this you can add a setTheme(isDark) action to the store and bind directly with :active and @update:active, no local computed needed.

src/pages/keyword-density/KeywordDensity.vue
```diff
@@ -1,17 +1,35 @@
...
+const keywordStore = useKeywordStore();
+
 const STATIC_PARENT_TEXT = 'Our Keyword Counter tool lets you count how many times keywords are repeated in any text, and also calculates the density of these keywords. The keyword density is the percentage of times a keyword appears in a text compared to the total number of words in that text. Simply write or paste your text here and hit "count".';
 
-const inputText = ref(STATIC_PARENT_TEXT);
+onMounted(() => {
+    if (!keywordStore.sharedInputText) {
+        keywordStore.sharedInputText = STATIC_PARENT_TEXT;
+    }
+});
```
sevvalkaplannn: initialize it inside the store you don't need onMounted in here when you initialize it on store

src/pages/keyword-density/KeywordDensity.vue
```diff
@@ -1,17 +1,35 @@
...
+import { ref, computed, onMounted } from 'vue';
+
+import { useKeywordStore } from '@/stores/keywordStore';
+import { MaTextarea, MaButton, MaCheckbox2 as MaCheckbox, MaNotification, MaEmpty } from '@mobileaction/action-kit';
+import { processKeywordDensity } from '@/utils/calculateDensity';
 import { AgGridVue } from 'ag-grid-vue3';
 
+const keywordStore = useKeywordStore();
+
 const STATIC_PARENT_TEXT = 'Our Keyword Counter tool lets you count how many times keywords are repeated in any text, and also calculates the density of these keywords. The keyword density is the percentage of times a keyword appears in a text compared to the total number of words in that text. Simply write or paste your text here and hit "count".';
```
sevvalkaplannn: you have it on keywordStore too, it is duplicated now, use it directly from store

src/stores/keywordStore.js
```diff
@@ -0,0 +1,18 @@
...
+export const useKeywordStore = defineStore('keyword', () => {
+    const sharedInputText = ref('');
+
+    const STATIC_PARENT_TEXT = 'Our Keyword Counter tool lets you count how many times keywords are repeated in any text, and also calculates the density of these keywords. The keyword density is the percentage of times a keyword appears in a text compared to the total number of words in that text. Simply write or paste your text here and hit "count".';
+
+    function initDensityText() {
+        if (!sharedInputText.value) {
+            sharedInputText.value = STATIC_PARENT_TEXT;
+        }
+    }
```
sevvalkaplannn: it is not used, you can remove it

src/stores/keywordStore.js
```diff
@@ -0,0 +1,18 @@
...
+
+    return {
+        sharedInputText
+    };
```
sevvalkaplannn, line 87: a store should expose actions instead, setInputText(text), clearInput(), resetDefaultText() instead of every component changing refs directly you can add them to store

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -13,13 +17,25 @@ const nGramOptions = Array.from({ length: MAX_NGRAM_SIZE }, (_, i) => ({
...
 const generatedKeywordNGrams = ref([]);
-const hasInput = computed(() => sourceDescriptionText.value.trim().length > 0);
+const hasInput = computed(() => keywordStore.sharedInputText.trim().length > 0);
 const hasGeneratedKeywords = computed(() => generatedKeywordNGrams.value.length > 0);
 
+const lastGeneratedState = ref({
+    text: null,
+    removeStopWords: null,
+    nGrams: null
+});
```
```suggestion
});

```

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -13,13 +17,25 @@ const nGramOptions = Array.from({ length: MAX_NGRAM_SIZE }, (_, i) => ({
...
+const hasStateChanged = computed(() => {
+    const currentNGrams = [...selectedNGrams.value].sort().join(',');
+    return keywordStore.sharedInputText !== lastGeneratedState.value.text ||
+           shouldRemoveStopWords.value !== lastGeneratedState.value.removeStopWords ||
+           currentNGrams !== lastGeneratedState.value.nGrams;
+});
+
```
sevvalkaplannn: this used in both generator and density I guess, we can do it on store, we need to handle that kind of common implementation on store

sevvalkaplannn (changes requested):

sevvalkaplannn (approved):
good work 🚀

PR #113, opened 2026-07-02
ONB-201: Implement keyword generator

emiraydin10:
Implemented the ONB-201 keyword generator inside src/pages/keyword-generator/KeywordGenerator.vue.

This PR adds:

- Text input for user-provided content
- Input cleaning before keyword generation
- Unique 1-gram, 2-gram, and 3-gram keyword generation
- Duplicate removal with Set
- Result sections for each n-gram type
- A small refactor to render keyword sections dynamically with v-for

Implementation details:

- The first commit adds the working ONB-201 functionality.
- The second commit refactors the result rendering by introducing a computed keywordSections array.
- This avoids repeating almost the same template block for 1-Gram, 2-Gram, and 3-Gram sections.

---
I addressed the review feedback.

Changes:
- Derived hasGenerated from generated keyword sections using computed.
- Added a short explanation for the Unicode regex.
- Verified with yarn lint and yarn build.

### DIFF
--- a/src/pages/keyword-generator/KeywordGenerator.vue
+++ b/src/pages/keyword-generator/KeywordGenerator.vue
@@ -1,5 +1,5 @@
 <script setup>
-import { ref } from 'vue';
+import { computed, ref } from 'vue';
 
 const userInput = ref('');
 const generatedKeywords = ref({
@@ -8,6 +8,20 @@ const generatedKeywords = ref({
     threeGram: [],
 });
 const hasGenerated = ref(false);
+const keywordSections = computed(() => [
+    {
+        title: '1-Gram',
+        keywords: generatedKeywords.value.oneGram,
+    },
+    {
+        title: '2-Gram',
+        keywords: generatedKeywords.value.twoGram,
+    },
+    {
+        title: '3-Gram',
+        keywords: generatedKeywords.value.threeGram,
+    },
+]);
 
 const cleanInput = (input) => input
     .toLowerCase()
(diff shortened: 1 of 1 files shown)


src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,114 @@
...
+import { computed, ref } from 'vue';
+
+// Store the input and generated keyword groups as reactive state.
+const userInput = ref('');
+const generatedKeywords = ref({
+    oneGram: [],
+    twoGram: [],
+    threeGram: [],
+});
+const hasGenerated = ref(false);
```
emrekaansurgun: The `hasGenerated` flag works, but consider whether you really need it as a separate piece of state. Since `generatedKeywords` starts with empty arrays, you could derive this information instead:

```js
const hasGenerated = computed(() =>
  keywordSections.value.some(s => s.keywords.length > 0)
);
```

This way there's one less thing to keep in sync, if you ever reset the keywords, you don't need to remember to also reset `hasGenerated`. This pattern is called "derived state" and it's a core principle in Vue: if something can be computed from existing state, prefer `computed` over a separate `ref`. That said, your current approach is explicit and easy to read, so it's not wrong!

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,114 @@
...
+        keywords: generatedKeywords.value.twoGram,
+    },
+    {
+        title: '3-Gram',
+        keywords: generatedKeywords.value.threeGram,
+    },
+]);
+
+// Normalize the input into lowercase words separated by single spaces.
+const cleanInput = (input) => input
```
emrekaansurgun, line 32: One small thing: the regex `/[^\p{L}\p{N}]+/gu` uses Unicode property escapes (`\p{L}` for any letter, `\p{N}` for any number). This is great for international text support. It's worth adding an inline comment explaining *why* you chose this over a simpler `/[^a-z0-9]+/g`, so future readers understand the intent:

```js
// \p{L} matches any Unicode letter, \p{N} any Unicode number (u flag enables Unicode mode).
// This ensures non-ASCII input (e.g. Turkish, Arabic) is handled correctly.
.replace(/[^\p{L}\p{N}]+/gu, ' ')
```

emrekaansurgun (approved):

src/pages/keyword-generator/KeywordGenerator.vue
emiraydin10: Addressed, thank you.

src/pages/keyword-generator/KeywordGenerator.vue
emiraydin10: Addressed, thank you.

emiraydin10:
I addressed the review feedback.

Changes:
- Derived hasGenerated from generated keyword sections using computed.
- Added a short explanation for the Unicode regex.
- Verified with yarn lint and yarn build.

PR #114, opened 2026-07-03
ONB-202: Improve keyword generator experience

emiraydin10:
## Summary

Implemented the ONB-202 UX improvements for the keyword generator and addressed review feedback with additional validation safeguards.

- Added multi-selection support for choosing which n-gram sizes should be displayed
- Extended n-gram generation support from 1-gram up to 10-gram
- Added removable unwanted words input with comma-separated parsing
- Replaced the plain textarea with ActionKit `MaTextarea`
- Replaced the unwanted words native input with ActionKit `MaInput`
- Rendered generated keywords as read-only `MaBadge` results
- Kept generated results synchronized when selected n-gram sizes change after generation
- Disabled Generate when no n-gram size is selected
- Disabled Generate when the main text input is empty or only whitespace
- Disabled Generate when the input does not contain enough words for all selected n-gram sizes
- Added helper messages for invalid generation states
- Replaced fragile child CSS selector usage with an explicit `.ma-keyword-group` class
- Fixed long keyword badge overflow for 6–10 gram results
- Added concise comments around review-related implementation decisions

## Review Feedback Updates

Addressed the review feedback by:

- Preventing generation when no n-gram size is selected
- Adding a safety guard in `generateKeywords` for invalid generation states
- Replacing `.ma-keyword-results > div` with the semantic `.ma-keyword-group` class
- Using ActionKit `MaInput` for the unwanted words field instead of a native input
- Keeping the current 1–10 gram generation approach because it prevents stale results when selected n-gram sizes change after generation

## Additional UX Improvements

While testing the same flow, I also handled related empty-result cases:

- Generate is now disabled when the input text is empty
- Generate is now disabled when the selected n-gram sizes cannot be produced from the current word count
- When multiple n-gram sizes are selected, the input must contain enough words for the largest selected n-gram size
- Helper messages now explain why Generate is disabled

Example cases covered:

- Empty input + selected n-gram size
- Text input + no selected n-gram size
- 1 word + selected 3-gram
- 5 words + selected 5-gram and 7-gram
- Valid input with enough words for all selected n-gram sizes

## Implementation Notes

- `MaBadge` is used instead of `MaTagInput` because generated keywords are read-only output, not editable user input.
- Selected n-gram sizes are used as the visibility source for rendered result sections.
- Generated results are kept in sync after selection changes to avoid stale or empty-looking sections.
- Generate is disabled unless the form can produce visible results for all selected n-gram sizes.
- Long n-gram badges wrap inside their own result section instead of overflowing into nearby columns.
- ActionKit badge styling is overridden only within `KeywordGenerator.vue` using scoped CSS and `:deep`, so global ActionKit styles are not changed.
- `.ma-keyword-group` is used instead of a direct child selector to keep the CSS more maintainable.

## Manual Checks

Manually tested the following cases after implementation and follow-up fixes:

- Default 1, 2, 3 n-gram selection
- Selecting and unselecting different n-gram sizes between 1 and 10
- Changing selected n-gram sizes after clicking Generate
- Inputs with repeated words to verify duplicate removal
- Inputs with unwanted words such as `is`, `a`, `an`, `the`
- Comma-separated unwanted words with spaces and different casing
- Empty input
- Empty n-gram selection
- Short input with large n-gram selections such as 8, 9, 10
- Multiple selected n-gram sizes where only some could be generated
- Long 6–10 gram badge output to verify there is no visual overflow

## Checks

- `git diff --check`
- `yarn lint`
- `yarn build`

Build passed with only existing dependency/chunk-size warnings.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "@mobileaction/action-kit": "^1.58.20",
-    "vue": "^3.4.29",
+    "vue": "^3.5.12",
     "vue-router": "^4.3.3"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "lint": "eslint . --ext .vue,.js,.jsx,.cjs,.mjs --fix --ignore-path .gitignore"
   },
   "dependencies": {
+    "@mobileaction/action-kit": "^1.58.20",
     "vue": "^3.4.29",
     "vue-router": "^4.3.3"
   },
--- a/src/main.js
+++ b/src/main.js
@@ -1,9 +1,12 @@
 import { createApp } from 'vue'
+import { ActionKit } from '@mobileaction/action-kit'
+import '@mobileaction/action-kit/dist/style.css'
 import App from './App.vue'
 import router from './router'
 
 const app = createApp(App)
 
+app.use(ActionKit)
 app.use(router)
 
 app.mount('#app')
--- a/src/pages/keyword-generator/KeywordGenerator.vue
+++ b/src/pages/keyword-generator/KeywordGenerator.vue
@@ -57,7 +57,7 @@ const generateKeywords = () => {
         : [];
 
     generatedKeywords.value = Object.fromEntries(
-        selectedGramSizes.value.map((gramSize) => [
+        gramSizeOptions.map((gramSize) => [
             gramSize,
             generateUniqueNGrams(words, gramSize),
         ]),
--- a/src/pages/keyword-generator/KeywordGenerator.vue
+++ b/src/pages/keyword-generator/KeywordGenerator.vue
@@ -2,10 +2,15 @@
 import { MaBadge, MaTextarea } from '@mobileaction/action-kit';
 import { computed, ref } from 'vue';
 
-// Store the input and generated keyword groups as reactive state.
 const userInput = ref('');
+
+// Keep the original ONB-201 result set visible by default while allowing 1–10 selection.
 const selectedGramSizes = ref([1, 2, 3]);
+
+// Seed the editable list with a short set of common words instead of imposing a large fixed stop-word dictionary.
 const unwantedWords = ref('is, a, an, the');
+
+// Define the supported range once so the selector and generation logic cannot drift apart.
 const gramSizeOptions = Array.from({ length: 10 }, (_, index) => index + 1);
 const generatedKeywords = ref({
     1: [],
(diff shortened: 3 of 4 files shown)


emrekaansurgun (commented):
Good work overall! The feature works as described and the code is well-structured. A few things to keep in mind for future improvements:

**1. Missing validation when no n-gram size is selected**
If the user unchecks all gram sizes and clicks "Generate", the button still fires and `generatedKeywords` gets populated silently, but nothing is shown because `hasGenerated` stays `false` (no section has keywords). Consider disabling the button or showing a warning when `selectedGramSizes` is empty. This prevents a confusing user experience.

**2. All 10 gram sizes are always generated, even unselected ones**
`generateKeywords` computes all 10 gram sizes every time regardless of what the user selected. You mentioned in your notes that this is intentional to avoid stale results when the selection changes, that's a valid reason. But if performance ever becomes a concern with large inputs, you could instead regenerate only when the selection changes (using a `watch`), or compute lazily. For now it's fine, just worth being aware of the trade-off.

**3. CSS selector `.ma-keyword-results > div` is fragile**
Using a bare `div` as a selector couples your styles to the markup structure. If you later wrap the content in another element, the styles silently break. Prefer a semantic class name like `.ma-keyword-group` on the div itself, it's more explicit and easier to maintain.

**4. Inconsistent use of ActionKit components**
The main text input uses `MaTextarea` (good!), but the "Unwanted words" field uses a plain native `<input>`. ActionKit has an `MaInput` component, prefer it here for visual and behavioral consistency across the form.

emrekaansurgun (approved):

emiraydin10:
Thanks for the detailed review.

I pushed an update addressing the feedback and also added a few related UX safeguards I noticed while testing the same flow.

Changes made based on the review:
- Disabled the Generate button when no n-gram size is selected.
- Added a safety guard in `generateKeywords` so generation cannot run without a selected n-gram size.
- Replaced the fragile `.ma-keyword-results > div` selector with an explicit `.ma-keyword-group` class to make the markup/style relation easier to maintain.
- Replaced the native unwanted words input with ActionKit `MaInput` for better consistency with the ActionKit textarea.

Additional UX improvements I added while testing:
- Disabled Generate when the main text input is empty or only whitespace.
- Added validation for cases where the current input does not have enough words for the selected n-gram sizes.
- When multiple n-gram sizes are selected, Generate now requires enough words for the largest selected n-gram size, so all selected sections can produce results.
- Added helper messages for these invalid states to make the form behavior clearer.
- Added short implementation comments around the validation and styling decisions for review clarity.

Checks:
- `git diff --check`
- `yarn lint`
- `yarn build`

Build passed.

PR #118, opened 2026-07-06
ONB-204: Implement keyword density page

emiraydin10:
## Summary
- Added a new `/keyword-density` page for ONB-204
- Passed static initial text from the parent page to the child component on first load
- Added editable ActionKit textarea and Count button
- Implemented keyword count and density calculation
- Rendered each keyword as its own row with Keyword, Count, and Density columns
- Replaced the plain HTML density table with sortable AG Grid Community table
- Added total character count, including whitespace
- Added Copy to clipboard support for density results
- Improved responsive layout for mobile, tablet, and desktop screens
- Migrated the project package manager from Yarn to pnpm based on the updated onboarding document
- Extracted keyword cleaning and n-gram generation into shared utility functions
- Moved default stop words into a dedicated constants file
- Replaced native n-gram checkboxes with ActionKit `MaSelect2` multi-select
- Addressed mentor/review feedback around character count, row styling, table colors, mobile table width, and stale layout issues

## Implementation Notes
- The child component copies the incoming `text` prop into local state so the textarea can be edited without mutating parent-owned data.
- Text is normalized before keyword counting so punctuation and extra spacing do not create invalid keywords.
- Density is calculated from total word count, not unique keyword count.
- Each keyword is represented by its own row to keep sorting and row data consistent.
- AG Grid Community is used for the density result table with sortable Keyword, Count, and Density columns.
- ActionKit components are used for the non-table UI controls.
- Keyword cleaning and n-gram generation are kept in shared utility functions for reuse.
- Default stop words are stored separately and used as the editable unwanted-words seed.
- Generated keyword outputs are still rendered with `MaBadge` because ActionKit does not expose a standalone read-only tag component. `MaTagInput` is editable input-oriented, so `MaBadge` is the safer read-only output component.

## Checks
- pnpm lint
- pnpm build

## Manual Checks
- Opened `/keyword-density`
- Verified initial text loads correctly
- Verified textarea remains editable after initial text is passed from the parent
- Verified Count button generates keyword density results
- Verified each keyword appears in its own table row
- Verified Keyword, Count, and Density sorting works in AG Grid
- Verified Copy to clipboard output
- Verified empty input shows a no-results state
- Verified total character count includes whitespace
- Verified table does not exceed screen width on mobile, tablet, and desktop widths
- Opened `/keyword-generator`
- Verified default selected n-gram sizes are 1, 2, and 3
- Verified ActionKit multi-select supports multiple n-gram selections
- Verified generated keyword results stay in sync with selected n-gram sizes
- Verified unwanted words still filter generated keywords correctly


src/constants/stopWords.js
```diff
@@ -0,0 +1,2 @@
+// Seed the editable unwanted-word field without imposing a fixed stop-word dictionary.
+export const DEFAULT_STOP_WORDS = 'is, a, an, the';
```
emrekaansurgun: DEFAULT_STOP_WORDS is a plain comma-separated string rather than an array. Since every consumer treats it as a list of words, storing it as an array (['is', 'a', 'an', 'the']) would be more type-safe and avoid relying on string parsing/splitting later.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -180,37 +151,52 @@ const generateKeywords = () => {
...
 </template>
+
+<style scoped>
+.keyword-tags.ma-tag-input-wrapper {
+    border: none !important;
+    background: transparent !important;
+    padding: 0 !important;
+    box-shadow: none !important;
+    min-height: 0 !important;
+}
```
emrekaansurgun: You should use :deep() instead of !important. 

example:
:deep(.your-class) {}

src/utils/keywordUtils.js
emrekaansurgun, line 1: Nice job extracting cleanInput and generateUniqueNGrams into keywordUtils.js, that's a good refactor for reuse with the new density page.

src/pages/keyword-density/KeywordDensity.vue
emrekaansurgun, line 1: Density is stored as a formatted string ("12.3%") and re-parsed with parseFloat just for sorting. Keeping density as a raw number in the row data and formatting it only for display (AG Grid valueFormatter) would avoid this round trip.

src/pages/keyword-density/KeywordDensity.vue
```diff
@@ -0,0 +1,183 @@
...
+                class="block w-full max-w-full box-border"
+                :rows="14"
+            />
+
+            <div class="flex flex-wrap items-center justify-between gap-4 mt-4">
+                <MaButton
+                    html-type="button"
+                    color="red"
+                    size="small"
+                    class="!border-red-700 !bg-red-700 !text-white hover:!border-red-800 hover:!bg-red-800"
```
emrekaansurgun: color="red" combined with several !important overrides for a primary action button looks redundant, and red is normally reserved for destructive/danger actions. Might be worth using the design system's primary/default color here instead of overriding it.

src/pages/keyword-density/KeywordDensity.vue
```diff
@@ -0,0 +1,183 @@
...
+    const rows = results.value.map(({ keywordText, count, density }) => (
+        `${keywordText}\t${count}\t${density}`
+    ));
+    const clipboardText = ['Keyword\tCount\tDensity', ...rows].join('\n');
+
+    try {
+        await navigator.clipboard.writeText(clipboardText);
+        copyStatus.value = 'Copied to clipboard.';
+    } catch {
+        copyStatus.value = 'Clipboard access is unavailable.';
```
emrekaansurgun, line 95: The catch block only sets a user-facing message but swallows the actual error. A console.error here would help debugging clipboard failures in specific browsers.

emrekaansurgun (approved):

PR #123, opened 2026-07-09
Qnb 201 kursat hurkan aydin

kursathurkanaydin:


src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,36 @@
...
+  return ngrams;
+})
 </script>
 <template>
     <div class="ma-keywords-generator">
         <div class="ma-header">
             <span>Keyword Generator</span>
         </div>
     </div>
+    <input v-model="userInput" placeholder="Type Something">
```
emrekaansurgun: Small structural thing: the closing `</div>` of `.ma-keywords-generator` is right above this, so the input/p/hr/keyword list end up as siblings instead of being inside the wrapper. Was that on purpose? If not, moving all of this back inside the wrapper div would keep the component's markup/styling consistent.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,36 @@
...
+    <p>Cleaned User Input: {{cleanedInput}}</p>
+    <hr>
+    //Displaying the generated keywords
```
emrekaansurgun: Heads up: `//` isn't valid inside a Vue template, so this line will actually render as literal text on the page instead of being a comment. Use `<!-- Displaying the generated keywords -->` for template comments.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,36 @@
...
+    <div>
+      <ul v-for="(nGram, index) in results" :key="index">
+        {{`${index + 1}-Gram`}}
+        <li v-for="(keyword, index) in nGram" :key="index">
```
emrekaansurgun: Nit: this inner `v-for` reuses `index` as the variable name, which shadows the outer `index` from the `nGram` loop above. It still works since Vue scopes it locally, but renaming it to something like `keywordIndex` would make the two loops easier to tell apart at a glance.

src/utils/CleanInput.js
```diff
@@ -0,0 +1,7 @@
+export function cleanInput(input) {
+    return input
```
emrekaansurgun, line 3: This works fine for how it's used now, but since it's a shared util, it'd be safer to guard against `input` being `null`/`undefined` (e.g. `if (!input) return '';`) so it doesn't throw if someone reuses this function elsewhere later.

src/utils/GenerateNGram.js
```diff
@@ -0,0 +1,8 @@
+export function generateNGram(input, n) {
```
emrekaansurgun, line 1: Try calling `generateNGram('', 1)` in your head: `''.split(' ')` returns `['']`, so the function ends up returning an n-gram containing an empty string. That means when the input box is empty, the list will render one empty bullet point. Might be worth returning `[]` early if `input` is empty.

emrekaansurgun (approved):

PR #125, opened 2026-07-09
Qnb 202 kursat hurkan aydin

kursathurkanaydin:


src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -2,36 +2,44 @@
...
+    <div class="ma-header">
+      <span>Keyword Generator</span>
     </div>
+    <MaTextInput v-model="userInput" placeholder="Enter text..."/>
+    <p>Cleaned User Input: {{ cleanedInput }}</p>
+    <hr>
+
+    <select v-model="selectedNGrams" multiple>
+      <option v-for="n in ngramLimit" :key="n" :value="n">{{ n }}-Gram</option>
+      /
```
emrekaansurgun: There's a stray "/" character sitting between </option> and </select>. It renders as plain text inside the dropdown and looks like a leftover from editing, worth removing before merging.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -2,36 +2,44 @@
 import {computed, ref} from "vue";
 import {cleanInput} from "@/utils/CleanInput.js";
 import {generateNGram} from "@/utils/GenerateNGram.js";
+import {MaTextInput, MaBadge} from "@mobileaction/action-kit";
 
-const ngramLimit = 3;
+const ngramLimit = ref(10);
```
emrekaansurgun: Since ngramLimit never changes anywhere in this component, it probably doesn't need to be a ref. A plain const like before works fine, refs are mainly useful when a value actually updates over time.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -2,36 +2,44 @@
...
-        </div>
+  <div class="ma-keywords-generator">
+    <div class="ma-header">
+      <span>Keyword Generator</span>
     </div>
+    <MaTextInput v-model="userInput" placeholder="Enter text..."/>
+    <p>Cleaned User Input: {{ cleanedInput }}</p>
+    <hr>
+
+    <select v-model="selectedNGrams" multiple>
```
emrekaansurgun: Nice addition letting the user pick which n-grams to see. Two things to check: right now nothing shows until the user manually selects something, is that on purpose? Also since you already switched to MaTextInput and MaBadge above, it'd be good to check if action-kit has a select component too, so we don't mix native and design system elements in the same form.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -2,36 +2,44 @@
...
+      <option v-for="n in ngramLimit" :key="n" :value="n">{{ n }}-Gram</option>
+      /
+    </select>
+
+    <!--        Displaying the generated keywords-->
+    <div>
+      <ul v-for="nGram in selectedNGrams" :key="nGram">
+        {{ `${nGram}-Gram` }}
+        <MaBadge variant="blue" v-for="(keyword, keywordIndex) in results[nGram - 1]" :key="keywordIndex">
```
emrekaansurgun: This works, but relying on nGram - 1 to find the right index can get confusing later if the data shape changes. Might be cleaner to keep the n value together with its results instead of recalculating the index here.

src/utils/GenerateNGram.js
```diff
@@ -1,9 +1,12 @@
+import {stopWords} from "@/utils/StopWords.js";
+
 export function generateNGram(input, n) {
-    if (!input || input.length < n) return []
+    if (!input || input.length === 0) return []
```
emrekaansurgun: Good simplification. One edge case to think about: since stop words are filtered now, cleanedKeywords can end up shorter than n even when input isn't empty. It won't break anything since the loop just won't run, but a quick check for that would make the intent clearer for anyone reading this later.

src/utils/StopWords.js
```diff
@@ -0,0 +1,36 @@
+
+export const stopWords = new Set([
```
emrekaansurgun, line 2: Nice touch keeping stop words in their own file, keeps GenerateNGram.js clean. Just noting the list is pretty short right now, might be worth expanding later if accuracy matters.

emrekaansurgun (approved):

PR #126, opened 2026-07-10
Qnb 203 kursat hurkan aydin

kursathurkanaydin:


src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -12,52 +13,96 @@ const cleanedInput = computed(() => cleanInput(userInput.value))
...
     nGram = `${i + 1}-Gram`;
     options.push({label: nGram, value: nGram})
   }
   return options;
 })
 
-//to generate n-gram keywords
-const results = computed(() => {
+//generate keywords when button is clicked
+const getKeywords = () => {
```
emrekaansurgun, line 24: Nice approach making the keyword generation trigger on a button click. One small suggestion though: you now have two functions, getKeywords which builds the object and generateKeywords which just assigns the result to results.value. You could merge these into a single function that computes the ngrams and sets results.value directly, that removes one extra layer of indirection and makes it easier to follow at a glance.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -12,52 +13,96 @@ const cleanedInput = computed(() => cleanInput(userInput.value))
...
   let nGram;
   for (let i = 0; i < ngramLimit; i++) {
     nGram = `${i + 1}-Gram`;
     ngrams[nGram] = generateNGram(cleanedInput.value, i + 1)
   }
   return ngrams;
-})
+}
+
+//to generate n-gram keywords
```
emrekaansurgun, line 34: This comment used to describe the computed block above it, but now it sits above just a ref declaration, which is a bit misleading since this line doesn't generate anything by itself. It would read better if you moved this comment down to the generateKeywords function, since that is where the actual generation happens now.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -12,52 +13,96 @@ const cleanedInput = computed(() => cleanInput(userInput.value))
...
+const generateKeywords = () => {
+  results.value = getKeywords();
+}
 
 //to keep selected n-gram options sorted
 watch(selectedNGrams, (newVal) => {
   selectedNGrams.value = newVal.sort();
 })
 
+
```
emrekaansurgun, line 46: There are two empty lines in a row here. It is a tiny detail, but keeping just a single blank line between blocks keeps the file consistent and easier to scan.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -12,52 +13,96 @@ const cleanedInput = computed(() => cleanInput(userInput.value))
...
+
 </script>
 
 <template>
-  <div class="ma-keywords-generator">
-    <div class="ma-header">
-      <span>Keyword Generator</span>
+  <div
+      class="ma-keywords-generator max-w-2xl mx-auto mt-8 p-6 bg-white rounded-xl shadow-sm border border-gray-100 flex flex-col gap-5">
```
emrekaansurgun: This class list is getting long on a single line. It still works fine, but as components grow it becomes harder to scan all the utility classes at once. You could break them across multiple lines or group related ones together, spacing and layout classes first, then colors, then borders, for example.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -12,52 +13,96 @@ const cleanedInput = computed(() => cleanInput(userInput.value))
 const nGramSelectOptions = computed(() => {
   const options = [];
   let nGram;
-  for(let i = 0; i < ngramLimit; i++) {
+  for (let i = 0; i < ngramLimit; i++) {
```
emrekaansurgun, line 16: Good catch adding the missing space after for. Small consistency fixes like this make the codebase easier to read, keep an eye out for these while you are in a file.

emrekaansurgun (commented):
Overall this is a solid step, the button trigger works and the Tailwind setup looks correct. I left a few small comments inline, mostly about code organization and a couple of style nits. Nothing blocking, just things worth keeping in mind for the next iteration. Good work overall.

emrekaansurgun (approved):

PR #128, opened 2026-07-10
Qnb 204 kursat hurkan aydin

kursathurkanaydin:


src/components/KeywordDensity.vue
```diff
@@ -0,0 +1,131 @@
+<script setup>
+import {ref, watch} from "vue";
+import {MaTextarea, MaButton} from "@mobileaction/action-kit";
+import {AgGridVue} from "ag-grid-vue3";
+import "../../node_modules/.pnpm/ag-grid-community@36.0.0/node_modules/ag-grid-community/styles/ag-grid.css";
```
emrekaansurgun: Importing CSS straight from a deep node_modules/.pnpm path is risky. This exact path depends on the installed version and how pnpm resolves nested dependencies, so it can break easily after a lockfile update or on a teammate's machine. Since ag-grid-community already exposes its styles as part of the package, you can import them directly, for example: import "ag-grid-community/styles/ag-grid.css". This is shorter and safe against dependency tree changes.

src/components/KeywordDensity.vue
```diff
@@ -0,0 +1,131 @@
...
+    () => props.text,
+    (newText) => {
+      inputText.value = newText
+    }
+)
+
+const getKeywordsCount = () => {
+  if (!inputText.value) return {}
+  const keywordsCount = {}
+  const keywords = inputText.value.split(' ')
```
emrekaansurgun: Splitting only by a single space can cause wrong counts if the user types multiple spaces, tabs, or line breaks, since those will create empty strings as "words". Using a regex like inputText.value.trim().split(/\s+/) would handle those cases better and give more accurate keyword counts.

src/components/KeywordDensity.vue
```diff
@@ -0,0 +1,131 @@
...
+
+<template>
+  <div class="mx-auto w-full min-w-0 max-w-6xl px-3 sm:px-4">
+    <h1 class="mb-4 text-lg font-semibold text-gray-800 sm:text-xl">
+      Keyword Count & Density
+    </h1>
+
+    <div class="grid min-w-0 grid-cols-1 gap-3 lg:grid-cols-[1.1fr_0.9fr] lg:items-start lg:gap-4">
+      <section class="flex min-h-[220px] min-w-0 flex-col gap-3 rounded-lg border border-gray-200 p-4 sm:p-5">
+        <MaTextarea
```
emrekaansurgun, line 99: This component receives a "text" prop from KeywordGenerator (the cleaned input), but it also renders its own separate textarea here. This means the user has to type or paste their text twice, once in the generator and once here, to see the density results. It would be simpler and less confusing to just use the "text" prop directly for the calculations and drop this extra textarea, unless there is a specific reason users need a different text for density than for keyword generation.

src/utils/GenerateKeywords.js
```diff
@@ -0,0 +1,11 @@
+import {generateNGram} from "@/utils/GenerateNGram.js";
+
+export const generateKeywords = (input, ngramLimit) => {
```
emrekaansurgun, line 3: Good move pulling this logic out of the component and into its own util function. It makes the code easier to test on its own and keeps KeywordGenerator.vue focused on the UI. One small note: nGram is declared once with let and reused across loop iterations. It works fine here, but declaring it with const inside the loop body is a bit safer and clearer, since each iteration gets a fresh value instead of reassigning an outer variable.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -21,40 +23,24 @@ const nGramSelectOptions = computed(() => {
...
-  }
-  return ngrams;
-}
-
-//to generate n-gram keywords
-const results = ref({});
-
-const generateKeywords = () => {
-  results.value = getKeywords();
+const handleGenerate = () => {
```
emrekaansurgun, line 25: Nice rename here. handleGenerate and generatedKeywords describe what they actually do much better than the old getKeywords/generateKeywords/results names, which were easy to mix up. Small naming detail that makes the component noticeably easier to follow, keep doing this in future PRs.

emrekaansurgun (commented):
Overall this is solid work, good job splitting logic into utils and improving naming compared to the previous version. I left a few comments about the CSS import path, text tokenization, the duplicate text input in KeywordDensity, and a couple of small style notes. None of these are blockers, but worth a look before merging.

src/components/KeywordDensity.vue
```diff
@@ -0,0 +1,131 @@
...
+            v-model="inputText"
+            placeholder=""
```
emrekaansurgun:
```suggestion
```

emrekaansurgun (approved):

PR #129, opened 2026-07-13
Onb 201 emre

Emre0ruc:


src/cleanupResources.js
yagizmumcu, line 1: If there is no change, please do not alter this file, i guess your ide or formatter does that

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,57 @@
 <script setup>
+import { cleanDescription } from "../../utils/CleanDescription";
+import {
+  oneGramGeneration,
+  twoGramGeneration,
+  threeGramGeneration,
+} from "../../utils/nGramGeneration";
+
+import { ref, computed } from "vue";
```
yagizmumcu:
```suggestion
import { ref, computed } from "vue";

import { cleanDescription } from "../../utils/CleanDescription";
import {
  oneGramGeneration,
  twoGramGeneration,
  threeGramGeneration,
} from "../../utils/nGramGeneration";

```
please sort these imports,

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,57 @@
...
+const description = ref("");
+const cleanedDescription = computed(() => cleanDescription(description.value));
```
yagizmumcu, line 9: Is there a specific reason to use as computed?

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,9 +1,57 @@
...
+
+const oneGramKeywords = computed(() =>
+  oneGramGeneration(cleanedDescription.value),
+);
+const twoGramKeywords = computed(() =>
+  twoGramGeneration(cleanedDescription.value),
+);
+const threeGramKeywords = computed(() =>
+  threeGramGeneration(cleanedDescription.value),
+);
```
yagizmumcu: Instead of defining three different computed methods, why not considering n count as parameter ?

src/utils/nGramGeneration.js
yagizmumcu, line 1: There is no need to export all of inside util, use componentwise methods inside keywordgenerator file

yagizmumcu (changes requested):

src/cleanupResources.js
Emre0ruc: Yes, formatter altered it.

src/pages/keyword-generator/KeywordGenerator.vue
Emre0ruc: I was going to call the function directly in {{ }} structure (I guess it is called as interpolation) , then AI suggested using computed to prevent redundant function calls during renders.

src/cleanupResources.js
Emre0ruc: Fixed it

src/pages/keyword-generator/KeywordGenerator.vue
Emre0ruc: Changed them and implemented only one function with count parameter

src/utils/nGramGeneration.js
Emre0ruc: Since I import only two functions from utils, I kept it same.

src/pages/keyword-generator/KeywordGenerator.vue
Emre0ruc: Sorted

src/utils/nGramGeneration.js
```diff
@@ -0,0 +1,16 @@
...
+  for (let i = 0; i < descriptionArray.length - (n - 1); i++) {
+    arrNGram.push(descriptionArray.slice(i, i + n).join(" "));
+  }
+  return removeDuplicateNGrams(arrNGram);
+}
+
+function removeDuplicateNGrams(ngrams) {
+  const uniqueNGrams = new Set(ngrams);
+  return [...uniqueNGrams];
+}
```
yagizmumcu:
```suggestion

function removeDuplicateNGrams(ngrams) {
  const uniqueNGrams = new Set(ngrams);
  return [...uniqueNGrams];
}

export function nGramGenerater(cleanedDescription, n) {
  const descriptionArray = cleanedDescription.split(" ");
  if (descriptionArray.length < n) {
    return [];
  }
  const arrNGram = [];
  for (let i = 0; i < descriptionArray.length - (n - 1); i++) {
    arrNGram.push(descriptionArray.slice(i, i + n).join(" "));
  }
  return removeDuplicateNGrams(arrNGram);
}
```
please fix the order

.gitignore
```diff
@@ -28,3 +28,5 @@ coverage
 *.sw?
 
 *.tsbuildinfo
+
+.prettierignore
```
yagizmumcu:
```suggestion
```
please do not alter gitignore,

yagizmumcu (changes requested):
Image
Please also style the content and increase the width and height of textarea,

Secondly, please add a button to navigate the keyword generator on home screen

yagizmumcu (approved):

PR #131, opened 2026-07-14
Qnb 205 kursat hurkan aydin

kursathurkanaydin:


src/components/KeywordDensity.vue
```diff
@@ -1,49 +1,13 @@
 <script setup>
-import {ref, watch} from "vue";
+import {computed, ref} from "vue";
 import {MaTextarea, MaButton} from "@mobileaction/action-kit";
 import {AgGridVue} from "ag-grid-vue3";
+import {useUserInputStore} from "@/stores/UserInput.js";
```
emrekaansurgun, line 5: Nice job replacing the props + watch pattern with a single computed! This removes the need to manually keep inputText in sync and is more idiomatic Vue.

One thing worth thinking about: by importing useUserInputStore directly inside this component, KeywordDensity is now tightly coupled to that specific store instead of receiving its input via a prop. That makes it harder to reuse this component elsewhere or test it in isolation. It's a valid choice if this component will always be tied to this store, just good to be aware of the trade-off.

src/pages/keyword-generator/KeywordGenerator.vue
```diff
@@ -1,35 +1,32 @@
...
+const selectedNGramsStore = useSelectedNGramsStore()
+const generatedKeywordsStore = useGeneratedKeywordsStore()
+const cleanedInput = computed(() => cleanInput(userInputStore.userInput))
+const selectedNGramsModel = computed(
+    {
+      get: () => selectedNGramsStore.selectedNGrams,
+      set: (value) => selectedNGramsStore.setSelectedNGrams(value),
+    }
+)
+const sampleInput = 'Quick brown fox jump over fox';
```
emrekaansurgun: Small nit: sampleInput is hardcoded here as a literal string. Consider moving example/sample data like this into a constants file so it's easier to find and change later, especially if more sample inputs get added.

src/stores/SelectedNGrams.js
```diff
@@ -0,0 +1,33 @@
...
+        const options = [];
+        for (let i = 0; i < ngramLimitStore.ngramLimit; i++) {
+            const nGram = `${i + 1}-Gram`;
+            options.push({label: nGram, value: i + 1})
+        }
+        return options;
+    })
+
+    const setSelectedNGrams = (newValues) => {
+        newValues = [...newValues].sort((a, b) => a - b)
```
emrekaansurgun, line 19: Nice catch changing the option value from the string nGram to the number i + 1, and sorting numerically here. The old version sorted strings, which would have produced wrong ordering once you had 10+ grams (e.g. "10-Gram" would come before "2-Gram" alphabetically). Good example of why default sort() needs a comparator for numbers.

src/stores/NGramLimit.js
```diff
@@ -0,0 +1,10 @@
+import {defineStore} from "pinia";
+import {ref} from "vue";
+
+export const useNGramLimit = defineStore('ngramLimit', () => {
```
emrekaansurgun: Small consistency point: all the other stores in this PR are named like useXxxStore (useUserInputStore, useSelectedNGramsStore, useGeneratedKeywordsStore), but this one is useNGramLimit. Renaming it to useNGramLimitStore would keep the naming convention consistent, so it's instantly recognizable as a store.

src/stores/UserInput.js
```diff
@@ -0,0 +1,7 @@
+import {defineStore} from 'pinia'
+import {ref} from "vue";
+
+export const useUserInputStore =
```
emrekaansurgun: This store is written in a different style than the others (implicit object return) while NGramLimit.js and GeneratedKeywords.js use an explicit function body with a return statement. Both work the same way, but mixing styles makes the codebase harder to scan. Worth sticking to one style for all setup stores, e.g. defineStore('userInput', () => { const userInput = ref(''); return { userInput } }).

src/utils/ComputeKeywordCounts.js
```diff
@@ -0,0 +1,16 @@
+
+export const computeKeywordCounts = (inputText) => {
```
emrekaansurgun, line 2: Good move extracting this out of the component into its own pure util function, it's now reusable and much easier to unit test independently from the Vue component. Since it no longer depends on component state, it'd be a great candidate for a couple of quick unit tests (e.g. empty input, repeated keywords) to lock in this behavior.

emrekaansurgun (commented):
Overall this is solid progress. Good use of Pinia to share state between KeywordGenerator and KeywordDensity, and good extraction of the density/count logic into reusable utils. I left a few comments inline, mostly small consistency/naming points and one architectural note about how KeywordDensity now gets its data directly from the store instead of via props. Nothing blocking, just things to think about for next steps. Great work overall!

emrekaansurgun (approved):

PR #134, opened 2026-07-16
Onb 206 kursat hurkan aydin

kursathurkanaydin:


cypress/e2e/keyword-generator.cy.js
```diff
@@ -0,0 +1,105 @@
...
+        cy.get('[data-cy="keyword-generator-input"] input')
+            .should('be.visible')
+            .type(inputText)
+
+        cy.get('[data-cy="keyword-generator-generate"]').click()
+
+        cy.get('[data-cy="keyword-generator-select"]').click()
+        cy.contains('1-Gram').should('be.visible').click()
+
+        cy.get('body').click(0, 0)
```
emrekaansurgun, line 18: Clicking on (0, 0) to close the dropdown works, but it's a bit hard to read and can be fragile if something else ever sits in that corner. Using cy.get('body').type('{esc}') is usually clearer and communicates the intent better: "close this dropdown".

cypress/e2e/keyword-generator.cy.js
```diff
@@ -0,0 +1,105 @@
...
+
+        cy.get('[data-cy="keyword-generator-generate"]').click()
+
+        cy.contains('1-Gram:').should('be.visible')
+        cy.contains('3-Gram:').should('be.visible')
+        cy.contains('5-Gram:').should('be.visible')
+        cy.contains('4-Gram').should('not.be.visible')
+    })
+
+    it('remove stop words', () => {
```
emrekaansurgun: Small style note: the other test names read like sentences starting with a verb, e.g. "enters text...", "filters generated...". Naming this one "removes stop words from generated keywords" would keep the style consistent and make the test report easier to scan.

src/components/KeywordDensity.vue
```diff
@@ -40,7 +39,7 @@ const defaultColDef = {
 const rowData = ref([])
 
 const computeDensityAndCountOfKeywords = () => {
-  const countMap = computeKeywordCounts(inputText.value)
+  const countMap = computeKeywordCounts(userInputStore.userInput)
```
emrekaansurgun, line 42: Nice cleanup here. The old `inputText` was just a computed alias for `userInputStore.userInput`, so it added an extra layer without real benefit. Using the store value directly is simpler and works fine because Pinia setup stores return reactive refs, so v-model can bind to them directly.

package.json
```diff
@@ -8,12 +8,15 @@
...
-    "lint": "eslint . --ext .vue,.js,.jsx,.cjs,.mjs --fix --ignore-path .gitignore"
+    "lint": "eslint . --ext .vue,.js,.jsx,.cjs,.mjs --fix --ignore-path .gitignore",
+    "test:e2e": "cypress run",
+    "test:e2e:open": "cypress open"
   },
   "dependencies": {
     "@mobileaction/action-kit": "^1.58.20",
     "ag-grid-community": "^36.0.0",
     "ag-grid-vue3": "^36.0.0",
+    "cypress": "^15.18.1",
```
emrekaansurgun: Since cypress is only used for testing and never runs in production, it should go under "devDependencies" instead of "dependencies". Keeping this separation helps keep the production bundle lean and makes it clear which packages are actually needed at runtime.

pnpm-workspace.yaml
```diff
@@ -1,4 +1,5 @@
 allowBuilds:
   core-js: false
+  cypress: false
```
emrekaansurgun: Question about this one: setting cypress to false here means pnpm won't run its install script, but Cypress needs that script to download its binary. Did you test a clean install (removing node_modules and reinstalling) to make sure `cypress open`/`run` still works? If it does, it would be good to note why in a comment, if not, this probably should be true.

emrekaansurgun (commented):
Good first step overall, the e2e tests cover the main flows and the data-cy attributes are named consistently. Left a few notes inline, mostly about reducing duplication in the test file and two small config questions (cypress in dependencies vs devDependencies, and the allowBuilds setting). Nothing blocking, just things worth discussing.

emrekaansurgun (approved):

